PR #8, opened 2019-01-17
[RFC] Indexing/iteration aspects on unbounded strings

raph-amiard:


briot:
When I developed GNATCOLL.Strings, I was really trying to work around some issues with the design of Unbounded_String, so this proposal would fix some of them. The fact that it is visibly tagged would let users use the more natural dot-notation to call methods.

In terms of implementation, GNATCOLL.Strings has some nice tricks to improve performance (generic accessors to get the String value instead of function returning string, optional copy-on-write, small-buffer optimization,...). All of these combine to make a package more flexible than unbounded_string, though the latter is nicer for new comers to the language. Perhaps the standard should think of building those simpler packages based on more general and configurable ones. As mentioned, this would apply to unbounded_string, but could also apply to the standard containers (to control size growth algorithm for vectors, for instance) to give more advanced more control.

considered/rfc-add-indexing-aspects-to-unbounded-strings.rst
```diff
@@ -0,0 +1,123 @@
...
+
+.. code-block:: ada
+
+   use Ada.Strings.Unbounded;
+
+   Length (My_String);
+
+versus
+
+.. code-block:: ada
```
yakobowski: For some reason, this block of code is not displayed when the .rst is rendered.

yakobowski:
I ilke this proposal a lot. We should try to understand the impact of what is mentioned in the 'drawbacks' section, by using modifying `Strings.Unbounded` and see whether we can successfully recompile the available codebases that use it.

sttaft:
Unbounded string need not be visibly tagged in the current standard, but it does need to be (internally) controlled in so far as automatically recovering storage.  So it would not be a major burden to make it visibly tagged.

PR #9, opened 2019-01-21
[RFC] Constant range indexing

raph-amiard:


briot:
+1 for making this a general facility applicable to user-defined types. I would have liked to use something similar for GNATCOLL.Strings.

But perhaps it would be better to think of this in a more general context where a range is an actual object that can be passed around.  C++ has just gained that feature (for prior-art), as well as languages like D and python.

As a use case, say I have a `Sum` function that receives an array of Float and adds them up.
If I want to use it on a subset of the array, I would pass `Arr (4..7)`. The same cannot be done with the Ada containers. I have a `Sum` for a vector, I have to write a second version of Sum that also receives the beginning and started indexes, and likely a third one that receives start and end as cursors. It would be nicer if I could write a `Sum` that receives a vector range as a parameter, and would automatically also accept a vector itself as a synonym for "full range".

Syntactically, ranges would be a different kind of types, built either by functions or via the ".." operator.

(just a rough idea, this would obviously need more thoughts and formalization)

sttaft:
It might make sense to call this "user-defined slicing" or "generalized slicing" since "range indexing" is not an Ada term.

QuentinOchem:
Note on the proposed function interface:

 function Slice (Self : T'Class; Low, High : Character) return T;

This is going to be very expensive as it involves a copy of T, and only absolutely required on type initialization, otherwise you already have the target constraints. I would suggest two interfaces instead, e.g.:

function Create_Slice (Self : T'Class; Low, High : Character) return T;
procedure Copy_Slice (Self : T'Class; Low, High : Character; Target : in out Some_Type);

So that a more efficient Copy_Slice can be used whenever possible.

PR #11, opened 2019-05-16
[RFC] Expression functions as generic formal function parameters

clairedross:


djthyrax:
I really like this proposal, having had the need of something similar a couple of days ago.

raph-amiard:
Thanks guys for your feedback ! @djthyrax @AdaDoom3 could you please upvote the OP ? it will make it easier for us to track support. TIA!

considered/rfc-expression-functions-as-default-for-generic-formal-function-parameters.rst
```diff
@@ -0,0 +1,168 @@
...
+
+generic
+   type T is private;
+package Stacks is
+   type Stack is private;
+   procedure Push (S : Stack; X : T);
+   ...
+
+However, if we use this generic to store access based structure, it may not do
+what we expect:
```
vser1, line 52: Is the repetition of `what we expect` on purpose?

yannickmoy: That's not what we expect indeed. ;-)
I've submitted a typo-fixing PR, thanks.

PR #13, opened 2019-05-28
[RFC] Unify end record syntax

AdaDoom3:


QuentinOchem:
I personally find this a well welcome change. This allows some syntax unification with other type structures, such as protected types and task types, as well as other elements that define architecture, including packages and subprograms.

raph-amiard:
I think this proposal is good. I also think it's pretty minor but goes into the right discussion wrt. improving syntax

bobduff:
I think Ada 202x should add just one new possibility:
either "end record Type_Name;" or "end Type_Name;".
The existing draft RM allows the former, which is consistent
with "end loop" syntax.  I somewhat prefer the latter,
and would go along with changing to that, but I see no
point in allowing both. ARG mostly preferred the former.

Allowing options whose usage is based on personal preference
results in multiple dialects of the language, making it
(slightly) harder for people to read each other's code.
And then everybody has to argue about which choice their
project's coding standards should require.

Minor point: the syntax for record_representation_clause
says "end [record] [local_name];".  It's slightly wrong
to use local_name here, because that invokes visibility
rules, which we don't want.  We want "identifier", and the
rule about "shall repeat", which bypasses the visibility
rules.

Hmm.  Now I see that the AI and the draft RM make the
same mistake.

yannickmoy:
I would all the way to making "end Type_Name;" the official Ada 202X syntax, and deprecate the syntax "end record;". That would solve the problem of language dialect, while keeping backward compatibility for the time being.

raph-amiard:
> I think Ada 202x should add just one new possibility:
> either "end record Type_Name;" or "end Type_Name;".

I agree. I just disagree with the ARG's choice in the matter.

> The existing draft RM allows the former, which is consistent
> with "end loop" syntax. 

... Except:

1. Records are part of declarations, which follow the "name ... end name" syntax. loops are statements.
2. Loop names, like block names, are optional, whereas record types are necessarily named. It means that forcing the inclusion of "loop" in closing loop names is justified by this.

> I somewhat prefer the latter,

So we agree :) In my opinion there is an objective argument of consistency besides bike-shedding though.

> and would go along with changing to that, but I see no
> point in allowing both.

Agreed.

> ARG mostly preferred the former.

As I said, I think it's not just a matter of preference, but also of consistency.

> Allowing options whose usage is based on personal preference
> results in multiple dialects of the language, making it
> (slightly) harder for people to read each other's code.
> And then everybody has to argue about which choice their
> project's coding standards should require.

I agree with that, even though my devil's advocate side needs to point out Ada is already full of those. end names being optional, `use` vs `use type` vs package renamings, etc.

raph-amiard:
> I would go all the way to making "end Type_Name;" the official Ada 202X syntax, and deprecate the syntax "end record;". That would solve the problem of language dialect, while keeping backward compatibility for the time being.

I agree fully with that, and it goes in the direction we're hinting towards in https://github.com/AdaCore/ada-spark-rfcs/pull/24.

@AdaDoom3 what do you think ?

QuentinOchem:
> I would go all the way to making "end Type_Name;" the official Ada 202X syntax, and deprecate the syntax "end record;". That would solve the problem of language dialect, while keeping backward compatibility for the time being.

I completely agree as well.

setton:
> > I would go all the way to making "end Type_Name;" the official Ada 202X syntax, and deprecate the syntax "end record;". That would solve the problem of language dialect, while keeping backward compatibility for the time being.
> 
> I agree fully with that, and it goes in the direction we're hinting towards in #24.
> 
> @AdaDoom3 what do you think ?

I agree as well. Always found "end record;" to be incongruously different from the rest.

clairedross:
I think adding new syntax is fine, but deprecating the only preexisting one does not seem too user friendly to me. End record is not so pretty, but it does not cause readability issues either (clash with other syntax for example.

correctimundo:
I might agree to go for the end Type_Name; syntax for Ada 202x as suggested by Bob, but I really think it would only be worthwhile if it *eliminates* inconsistency on some level, not just reduce it or worse reduce inconsistency in one aspect, but introduce further inconsistency in another aspect. In particular, I am thinking of the case ... end case syntax for variant record. I am fine if we are only concerned with eliminating inconsistency with regard to declarations, rather than with statements, but  the case ... end case variant syntax is part of a declaration, and even more important, part of a record declaration. I think people who write variant record declarations will want to write "end record" after writing "end case", and will find it confusing that we will have introduced yet another inconsistency, especially after so many years of getting used to the end record syntax. For that reason, I do not think I would support deprecating "end record" syntax, but could see it being an optional alternate form. I agree with Bob that it is best to avoid creating "dialects" of Ada, which is another reason why I am less excited about this proposed change, but I think could still be viewed as a slight improvement if we were to agree to not deprecate the "end record" form.

yannickmoy:
@correctimundo this is really a property of all other top-level declarations (tasks, protected types/objects, subprograms, packages) that they finish with "end [identifier];" thus this proposal simply restores some uniformity. This is different from other block constructs. Your objections concerning habits to change is valid, it's really a matter of whether we can or not cleanup the language for easier use in the future, or if we're bound to all past choices including for future language versions. I think that making changes and marking previous choices as obsolete but supported gives the best option.

correctimundo:
> @correctimundo this is really a property of all other top-level declarations (tasks, protected types/objects, subprograms, packages) that they finish with "end [identifier];" thus this proposal simply restores some uniformity. This is different from other block constructs. Your objections concerning habits to change is valid, it's really a matter of whether we can or not cleanup the language for easier use in the future, or if we're bound to all past choices including for future language versions. I think that making changes and marking previous choices as obsolete but supported gives the best option.

I understand the inconsistency against all other top-level declarations, but my complaint is against the inconsistency we'd be creating within the syntax for records itself, with respect to variant records.
I suspect that the variant record syntax; case .. end case is the reason why the original language designers chose the record syntax we currently have. If were to change the record .. end record syntax, then I'd like to see something similar done with the case ... end case syntax, if possible, to completely remove inconsistencies. Perhaps some sort of case expression-like syntax could be applied, since we now do have case expressions?
e.g.
     type Foo (Has_Bar : Boolean := True) is
          record
               case Has_Bar is
                   when True =>
                        Bar : Integer;
                   when False =>
                        null;
               end Has_Bar;
          end Foo;

This has the advantage over current syntax that for nested case statements, there is improved readability because the end <identifier> shows more clearly which nested case is being completed.

or we could consider that the case part doesn't have an identifier, and so an anonymous syntax could be applied, such as;

     type Foo (Has_Bar : Boolean := True) is
          record
               case Has_Bar is
                   when True =>
                        Bar : Integer;
                   when False =>
                        null;
          end Foo;

The "end case" doesn't really add a lot to the declaration, and could be considered as mostly noise.

yannickmoy:
@correctimundo given that variant parts can be nested in record type definitions, they do need a closing marker. And these markers are very similar to similar "end loop" / "end if" / "end case" in subprograms, so I'm not convinced that the change you propose is beneficial.

correctimundo:
> @correctimundo given that variant parts can be nested in record type definitions, they do need a closing marker. And these markers are very similar to similar "end loop" / "end if" / "end case" in subprograms, so I'm not convinced that the change you propose is beneficial.

I'm not convinced either, but I'm also not convinced that this proposed change is beneficial either.

I never seem to have trouble remembering to use the end <keyword> syntax for all the places it is used in Ada, including end record, but I sometimes have to think twice about whether to leave the keyword out in places where it is not allowed. e.g.  Is it end package Foo, or end Foo?
In fact, I just had to look it up to convince myself. So I think removing "record" from end record would be a step in the wrong direction, especially since I think it would be confusing having end case; followed by end identifier;

I think a better option in my opinion likely would be to fix the places that don't allow end keyword to at least allow it optionally. Then you have consistency, and this seems to be the reasoning behind the ARG choice.

It's interesting that for enumerations, Ada chose parenthesis as markers rather than define an enumeration keyword, as in;
    type Colors is
         enumeration
               Red,
               Blue,
               Green
        end enumeration;

We now have other syntax forms where parens are optional, such as expression functions vs regular functions.
Maybe it is worth considering whether to allow parens for record definitions, in a style similar to enumeration syntax and perhaps allow commas instead of semicolons for separators, similar to enumerations, where the last comma can be omitted.

type Foo is
   (A : Integer,
    B : Float);

Seems to be easy enough to understand for the reader. It also seems to be less clunky for record extensions, e.g.

type Bar is new Foo with (C : Character, D : Home_Address);

For simple record structures, I find this a lot nicer to read and write, in my opinion. For large record structures that span multiple pages, I would still prefer to use the syntax as defined in the AI, because when you are viewing the end of the record structure, you can tell that it is a record (as opposed to say a protected type definition), and you can tell which instance of record declaration is being completed.

correctimundo:
> but I'm also not convinced that this proposed change is beneficial either.

Actually, I am now convinced that this proposed change is more harmful than beneficial, because in addition to the inconsistency with variant record syntax, it doesn't seem to work well with record extensions. 

    type Foo is new Bar with
        record
            X : Integer;
        end Foo;

Here we expect the end marker, "end Foo" to tie in the the start marker, which is "record", but it now appears to tie in with the type name, which is outside the syntactic element of the record extension.

This seems broken or at least very inconsistent with the rest of the Ada syntax, in my view.

My suggestion to allow the use of parens in place of record ... end record still appeals to me, however, as it does seem to work nicely with record extension syntax, if we are still looking for ways to potentially improve the record syntax. Otherwise, I'm fine with the ARG choice to optionally allow the type identifier after "end record".

FredPraca:
> Here we expect the end marker, "end Foo" to tie in the the start marker, which is "record", but it now appears to tie in with the type name, which is outside the syntactic element of the record extension.

For me, _end Foo_ is more interesting than _end record_ in this case because the important thing is the type, not the fact that it is a record. Sometimes, when you have a big record such as a network protocol description which does not fit in your display, knowing that you're at the end of a record is less important that knowing which record type it is.

The only thing I thought that might be confusing is when you have a representation clause in addition to the record definition. In this case, we would have _end Record_Type_ twice but that's already the case with the current _end record_. 

In your case, I suspect that expecting _end record_ is only a consequence of the alignment of _record_ with _end record_ and by the way, only a matter of format.

yakobowski:
@AdaDoom3 can you update this issue regarding the discussions that took place last year at the ARG?

raph-amiard:
@AdaDoom3 It seems to me that this RFC has been killed by the last round of discussions. I'll let you add clarifications to it if needed.

PR #18, opened 2019-06-14
[RFC] Parallelism as a library

raph-amiard:
Direct link to file https://github.com/AdaCore/ada-spark-rfcs/blob/topic/rfc-parallelism-as-a-library/considered/rfc-parallelism-as-a-library.rst

### DIFF
--- a/considered/rfc-parallelism-as-a-library.rst
+++ b/considered/rfc-parallelism-as-a-library.rst
@@ -332,8 +332,12 @@ parallel ARG AI.
             Mid : constant Positive := S'First + S'Length/2 - 1;
           begin
             Parallel_Do ((
-              procedure Res := Search (S (S'First .. Mid), Char) end;
-              procedure Res := Search (S (Mid + 1 .. S'Last), Char) end;
+              procedure is begin
+                 Res := Search (S (S'First .. Mid), Char);
+              end,
+              procedure is begin
+                 Res := Search (S (Mid + 1 .. S'Last), Char);
+              end
             ));
             return Res;
           end;
--- a/considered/rfc-parallelism-as-a-library.rst
+++ b/considered/rfc-parallelism-as-a-library.rst
@@ -184,7 +184,7 @@ current Ada:
       for I in 1 .. Num_Terms loop
          F (I) := (-1.0) ** (I - 1) * X ** (2 * I - 1) / Float (Fact(2 * I - 1));
       end loop;
-      return Red (F, 0.0);
+      return Red (0.0, F);
    end Sin;
 
 Or, if we imagine we already have map expressions:
@@ -267,7 +267,7 @@ Lastly, using inference of actual generic actuals using call actuals
           (for in in 1 .. Num_Terms =>
             (-1.0) ** (I - 1) * X ** (2 * I - 1) / Float (Fact(2 * I - 1)))));
 
-Here, the only generic actual we have to specify is \``, because:
+Here, the only generic actual we have to specify is \`Fn`, because:
 
 -  All array type parameters are infered from the ``Self`` actual
    parameter. ``Self`` allows us to deduce the type of the
--- a/considered/rfc-parallelism-as-a-library.rst
+++ b/considered/rfc-parallelism-as-a-library.rst
@@ -1,0 +1,521 @@
+Parallelism as a library
+========================
+
+* Feature Name: parrallelism_as_library
+* Start Date: 2019-06-14
+* RFC PR: (leave this empty) 
+* RFC Issue: (leave this empty)
+
+.. IMPORTANT:: 
+
+   The following document is a meta RFC, describing a general
+   approach using specific features. As such, it will suffer from several
+   caveats.
+
+   * It uses syntax that is not formally defined
+   * It references RFCs that are not yet written
+
+   The goal of this meta RFC is to create discussion and gather feedback on a
+   high level approach. As such it is considered that those caveats are
+   acceptable, and can/will be fixed when appropriate.
+
+Summary
+-------
+
+The Ada 2020 standard introduces two new aspects, ``'Reduce`` and
+``'Parallel_Reduce``. Those aspects allow possibly parallel
+`folding <https://en.wikipedia.org/wiki/Fold_(higher-order_function)>`__
+of collections.
+
+This RFC proposes to instead use generic functions for reduction, and
+introduces a series of enhancements to Ada generics which purpose is to
+make the call to those generic functions as expressive as the use of the
+reduce attribute.
+
+The Ada 2020 standard also introduces more general purpose 
+`parallel operations
+<http://www.ada-auth.org/cgi-bin/cvsweb.cgi/ai12s/ai12-0119-1.txt>`_. 
+
+This RFC also proposes to replace this heavy language level support by
+orthogonal improvements to the language, and eventual library support.
+
+Motivation
+----------
+
+This RFC proposes to instead use generic functions for reduction, and
+introduces a series of enhancements to Ada generics which purpose is to
+make the call to those generic functions as expressive as the use of the
+reduce attribute. This way, instead of introducing more special cases in
+the implementation of Ada compilers, we will:
+
+1. Introduce more general enhancements in the language that allow user
+   to implement their own powerful abstractions.
+2. Fix deficiencies of Ada generics that users have been complaining
+   about for a long time.
+
+This will also imply being largely more coherent, internally (reduce is
+a function like any other) as well as with regards to other languages,
+where reduce is typically implemented as a function.
+
+This would ultimately allow people to write the following example (using
+proposed Ada 2020 semantics)
+
+.. code-block:: ada
+
+   function Sin (X : Float; Num_Terms : Positive := 5) return Float 
+   is 
+     ([for I in 1..Num_Terms => 
+        (-1.0)**(I-1) * X**(2*I-1)/Float(Fact(2*I-1))]'Reduce("+", 0.0));
+
+This way:
+
+.. code:: ada
+
+   type Float_Array is array (Positive range <>) of Float;
+
+   function Sin (X : Float; Num_Terms : Positive := 5) return Float 
+   is 
+     (Reduce (Fn => "+")
+       (Float_Array'
+         ([for I in 1..Num_Terms => (-1.0)**(I-1) * X**(2*I-1)/Float(Fact(2*I-1))])))
+
+..
+
+   Note: Since reduce is a regular function, and not a magical
+   attribute, it’ll need to work on an object of a non ambiguous type,
+   hence the qualification above.
+
+Regarding parallel loops and blocks, our aim is to allow people to write the
+following proposed Ada 2020 code:
+
+.. code:: ada
+
+     A, B, C : Int_Array;
+
+     parallel for I in A'Range loop
+          A (I) := B (I) + C (I);
+     end loop
+
+into
+
+.. code:: ada
+
+     A, B, C : Int_Array;
+
+     Parallel (A'Range) (
+        procedure (I) is begin
+           A (I) := B (I) + C (I);
+        end
+     );
+
+An additional benefit is that fine tuning parametrization of parallel
+iteration can use simple subprogram parameters, instead of needing language
+extensions via pragmas or aspects.
+
+For example, scheduling/chunking, in the above example, can directly use a
+parameter to parallel iterate, with the following structure (to take back
+OpenMP's scheduling params):
+
+.. code:: ada
+
+     type Scheduling_Kind is (Static, Dynamic, Guided, Auto);
+     type Scheduling (Kind : Scheduling_Kind := Static) is record
+        case Kind is
+           when Static | Dynamic | Guided : Chunk_Size : Chunk_Size_Type
+             := No_Default;
+           when Auto => null;
+        end case;
+     end record;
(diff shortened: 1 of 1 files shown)


briot:
As a big user of generics, even the step 2 for inference of dependence types is a major step forward.
It can be used in a lot of contexts, including the standard libraries `Unchecked_Deallocation` or `Float_IO`, where currently users have to provide both the designated type and the access type when only the latter is really relevant. When writing a containers libraries like the traits-containers, it simplifies significantly the instantiations that the user has to do.

The implicit instantiation example has an extra semicolon in the expression function which I believe wasn't meant to be there. Though it would simplify the usage a lot, implicit instantiation also makes compiler error messages much harder to understand, since there is no named entity to attach the error message to. As given the example looks more like "inline instantiation" or "anonymous instantiation" than "implicit instantiation" though, since the user still have full control of which actual parameters are given to the generic. Such a feature can likely be quite easily experiment via a libadalang preprocessor.

These inline instantiations will put some pressure on the compiler to implement shared generics when possible (some compilers already do I believe, but not GNAT), because otherwise we might end up with very large code sides.

I like the fourth step (inference of actual parameters. Given the precise description you give, it looks like this is something libadalang would be quite comfortable implementing, though it might be a lot of work for other compilers ? Again, the likely difficulty is precision of error messages if we want to avoid the mess that C++ makes in similar contexts. Likely we'll need multi-line error messages that explain the chain of reasoning that the compiler applied/tried to apply.

I like the proposed syntax for anonymous functions. Very Ada-like and readable. I am less fascinated by the anonymous procedure proposal made in the parallel loop example that follows.

On the whole, I agree with the general approach that anything that can be implemented in libraries should be done there (and of course where possible provide a reference implementation). Focusing on the lower-level needs is way richer and can be reused in other context by users. We have little use for `Reduce` (and even less for a parallel Reduce), but any improvement to the generics is most welcome.

raph-amiard:
Thank you @briot for your comments

Regarding

> I like the fourth step (inference of actual parameters. Given the precise description you give, it looks like this is something libadalang would be quite comfortable implementing, though it might be a lot of work for other compilers ? Again, the likely difficulty is precision of error messages if we want to avoid the mess that C++ makes in similar contexts. Likely we'll need multi-line error messages that explain the chain of reasoning that the compiler applied/tried to apply.

The problem with C++ is that it has no contract model for generic bodies, ie in the following code:

```cpp
#include <iostream>

template <typename T> int foo(T inst) {
    return inst.bar();
}

class Pouet {
    public:
        int bar() { return 12; }
};

int main() {
    Pouet pouet_inst;
    std::cout << foo(pouet_inst) << std::endl;
}
```

The compiler will check for every instantiation whether there is a `bar()` member with the right signature or not.

Here, we keep the regular Ada contract model, so we don't have the complexity of C++. What we have is akin to the type inference needed to resolve overloads, so should not be much more complicated than what the compiler can already do.

I will of course detail that when writing the corresponding RFC, but basically, to resolve the following code:

```ada
   procedure Test is
      function Bar return Integer is (12);
      function Bar return Float is (12.0);
      function Bar return String is ("12");

      procedure Foo (A : Integer) is null;
   begin
      Foo (Bar);
   end Test;
```

The compiler will need to construct the set of possible types for the expression `Bar`, and match that against the expected type(s) in the call to `Foo`. If there is an ambiguity (eg. several matches) then it will say so, and possibly list the possible candidates.

Similarly, in the following theoretical example:

```ada
   procedure Test is
      generic type T is private; function Bar return T;
      procedure Foo (A : Integer) is null;
   begin
      Foo (Bar);
   end Test;
```

The call to bar would match against *any* possible type(s) in the call to `Foo`. If there is an ambiguity, then the compiler could say so, and list the possible matches too.

> These inline instantiations will put some pressure on the compiler to implement shared generics when possible (some compilers already do I believe, but not GNAT), because otherwise we might end up with very large code sides.

Yes, that's true. However, that can depend on the expansion model chosen for the generated instantiations. I believe there is a simple scheme in which we will share "as much as possible", without big modifications to GNAT.

considered/rfc-parallelism-as-a-library.rst
```diff
@@ -0,0 +1,525 @@
...
+where reduce is typically implemented as a function.
+
+This would ultimately allow people to write the following example (using
+proposed Ada 2020 semantics)
+
+.. code-block:: ada
+
+   function Sin (X : Float; Num_Terms : Positive := 5) return Float
+   is
+     ([for I in 1..Num_Terms =>
```
yannickmoy: why use square brackets here? you could just use the parentheses which are more usual

considered/rfc-parallelism-as-a-library.rst
```diff
@@ -0,0 +1,525 @@
...
+Or, if we imagine we already have map expressions:
+
+.. code:: ada
+
+   function Sin (X : Float; Num_Terms : Positive := 5) return Float is
+      function Red is new Array_Reduce (Positive, Float, Float_Array, Float, 0.0, "+");
+   begin
+      return Red
+        (0.0, Float_Array'
+          (for in in 1 .. Num_Terms =>
```
yannickmoy: typo: in in

considered/rfc-parallelism-as-a-library.rst
```diff
@@ -0,0 +1,525 @@
...
+
+.. code:: ada
+
+   function Sin (X : Float; Num_Terms : Positive := 5) return Float is
+      --  Index and element types are automatically deduced
+      function Red is new Array_Reduce  (<>, <>, Float_Array, Float, "+");
+   begin
+      return Red
+        (0.0, Float_Array'
+          (for in in 1 .. Num_Terms =>
```
yannickmoy: typo: in in

considered/rfc-parallelism-as-a-library.rst
```diff
@@ -0,0 +1,525 @@
...
+to the initial intent. Using implicit instantiation of generic functions
+`(see RFC here (TODO)) <https://TODO>`__, we would be able to write the
+following:
+
+.. code:: ada
+
+   function Sin (X : Float; Num_Terms : Positive := 5) return Float is
+     (Array_Reduce (<>, <>, Float_Array, Float, 0.0, "+");
+        (0.0, Float_Array'
+          (for in in 1 .. Num_Terms =>
```
yannickmoy: typo: in in

considered/rfc-parallelism-as-a-library.rst
```diff
@@ -0,0 +1,525 @@
...
+
+Lastly, using inference of actual generic actuals using call actuals
+`(see RFC here (TODO)) <https://TODO>`__, we can express the above as:
+
+.. code:: ada
+
+   function Sin (X : Float; Num_Terms : Positive := 5) return Float is
+     (Array_Reduce (Fn => "+")
+        (0.0, Float_Array'
+          (for in in 1 .. Num_Terms =>
```
yannickmoy: typo: in in

considered/rfc-parallelism-as-a-library.rst
```diff
@@ -0,0 +1,525 @@
...
+       Inst : A := new Integer'(12);
+   begin
+       Unchecked_Deallocation (Inst);
+   end Main;
+
+.. code:: ada
+
+   function Sin (X : Float; Num_Terms : Positive := 5) return Float is
+       Terms : Float_Array :=
+         (for in in 1 .. Num_Terms =>
```
yannickmoy: typo: in in

considered/rfc-parallelism-as-a-library.rst
```diff
@@ -0,0 +1,525 @@
...
+Further, using anonymous subprograms, we can express any parallel construct
+(parallel blocks or loops) in a pretty expressive way, where the example
+written in the rationale is made possible:
+
+.. code:: ada
+
+     A, B, C : Int_Array;
+
+     Parallel (A'Range, Scheduling'(Guided, 4)) (
+        procedure (I)
```
yannickmoy: you probably still want to provide a type for I in the prototype of the anonymous procedure here. and use "is" and "begin" keywords.

considered/rfc-parallelism-as-a-library.rst
```diff
@@ -0,0 +1,525 @@
...
+Where every instantiation is then used only once. Being able to avoid
+such unnecessary boilerplate seems like a worthy enough goal.
+
+Subprograms taking subprograms as arguments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The runtime already contains ``Iterate`` procedures, that takes callbacks as
+arguments. In the same spirit as for parallelism, the ARG has `proposed a
+special case syntax
+<http://www.ada-auth.org/cgi-bin/cvsweb.cgi/ai12s/ai12-0187-1.txt?rev=1.14&raw=N>`_
```
yannickmoy: this AI points to "Stable properties of abstract data types". I don't see how it relates to the special syntax for Iterate?

QuentinOchem:
As a meta comment on this meta rfc, we probably don't really need an rfc for parallelism as a library at this stage. The features needed to allow this are the ones of interest - once there, this is something that could be experimented on the side. In particular, the current proposal doesn't work well with GPU, but this is something that can be studied separately.

Glacia:
My problem with this RFC is that it feels like it's "disguised" to be about parallelism library, but in reality it's about introducing quite a lot of new features and parallelism library is just used as an example.  I think this parallelism functionality can be implemented with Ada 2012, so in my opinion it would be more appropriate to drop all of the new functionality introduced here and rewrite this RFC in Ada 2012 or drop this parallelism library disguise and rename this RFC to something like "general generics improvement" (i'm not good with names, but i hope you get the idea).

As for parallelism as a library: I agree that it would be better to have parallelism functionality as a library, mostly because features introduced (i.e. Reduce, parallel blocks/loops) seem pretty useless to me. At least if they're implemented as a library it would be easier to change them later.

raph-amiard:
Hi @Glacia, your comment seems totally fair to me, and is in line with what @QuentinOchem said. That RFC is kind of incomplete without the corresponding RFCs, that all correspond to improvements in the generic model, and can stand in their own right.

I will just say that the aim is not to "disguise" those improvements as parallel features, but rather to show how more general improvements to the generic model will allow just as much expressivity for parallel features, with more flexibility (due to the fact that we won't bake the features in the language), and more general benefits.

AdaDoom3:
I don't have too much to add that hasn't been said already. I would consider the parallel topic - like others here - not particularly relevant to what I feel is the core proposal. Namely implicit generics and anonymous subprograms.

Even though there will be implementation difficulty and diagnostics may be hampered a bit - I highly support these two features and look forward to their individual RFS : )

yakobowski:
I would _love_ to see the improvements you propose for generics, as the new concise syntax would be a huge plus. Likewise for anonymous subprograms, that I actually find even more important (probably because I come with a functional language mindset).

Regarding the topic of the PR, I very much agree with Manu's rephrasing:
> I agree with the general approach that anything that can be implemented in libraries should be done there

My main worry -- assuming that this get adopted and that we manage to implement it -- is that we may get (terrible) C++-template-resolution-like messages. You already answered this concern partly in https://github.com/AdaCore/ada-spark-rfcs/pull/18#issuecomment-502161299 though, so nothing new here :smile:

mosteo:
I'm going to throw a bit of discord here, but one of the things I love in Ada is that tasking is integral to the language. In that sense, I like that parallelism will get (will it?) specific keywords/syntax.

The improvements to generics look great, and the lack of type inference has always made me feel a bit silly telling the compiler things that cannot be otherwise. 

But, in regard to generics reuse, I guess there are two distinct aspects to it: reuse of a same instantiation (implicit with same arguments), and reuse of bodies for all instances (with different arguments) (which GNAT I guess doesn't plan to do anytime soon?). Just anecdotal evidence here, but I develop a generics-intensive library and compile times are through the roof... and touching a semicolon in a body is enough to retrigger almost everything.

considered/rfc-parallelism-as-a-library.rst
mgrojo: I think the correct link is: http://www.ada-auth.org/cgi-bin/cvsweb.cgi/ai12s/ai12-0009-1.txt?rev=1.10

sttaft: Actually, I believe the reference should be to the "loop body as procedure": 
   http://www.ada-auth.org/cgi-bin/cvsweb.cgi/ai12s/ai12-0189-1.txt?rev=1.14
-Tuck

sttaft:
One of the big challenges for using a library to implement parallelism is to provide any sort of compile-time checking.  However, the Parallel_Iterator aspect that has been proposed as part of the "Loop-Body" procedure feature might be generalized to work on any user-defined procedure that produces parallel or pseudo-parallel threads of control, provided that you also have some kind of Global aspect for declaring side-effects of subprograms so the checking can be performed at compile time, rather than as some separate tool that requires a full closure of the program.  In my experience, compile-time checking for data races is extremely important to making parallel programming productive -- just as much as compile-time type checking is important whenever a program gets large.

Lucretia:
Why not introduce ```lambda``` and a rename Like ```Pi```, ```λ```? It would make searching for anonymous functions easier.

raph-amiard:
We're closing this RFC because it has sparked discussion, and since it's a meta RFC and cannot be accepted as such, we see no value keeping it open for now.

PR #20, opened 2019-06-14
[RFC] Relaxed initialization for SPARK

clairedross:
Link to rich view:

https://github.com/AdaCore/ada-spark-rfcs/blob/topic/spark-relaxed-init/considered/rfc-spark-relaxed-init.rst

### DIFF
--- a/considered/rfc-spark-relaxed-init.rst
+++ b/considered/rfc-spark-relaxed-init.rst
@@ -23,7 +23,9 @@ by SPARK in general, but to add a new annotation, named Relaxed_Initialization,
 to allow relaxing it on a case by case basis. This annotation would come as a
 new aspect which can be used to exempt objects from the data initialization
 policy of SPARK:
-   
+
+.. code:: ada
+
    V : T with Relaxed_Initialization;
 
    function F (X : in out T; Y : out T; Z : in T) return T
@@ -38,6 +40,8 @@ initialized subcomponent in an object.
 In addition, to be able to speak about initialization of these objects in
 contracts, we would introduce a new 'Initialized attribute which would be True
 if the variable has been initialized:
+
+.. code:: ada
 
    type T is record
      F, G : Integer;
@@ -55,9 +59,11 @@ forbidden by the strict data initialization policy of SPARK. For example, it
 would allow having a Read procedure which would only initialize its Data out
 parameter if no errors occurred during the read, as modeled by an Error flag:
 
-procedure Read_T (Data : out T; Error : out Boolean) with
-     Relaxed_Initialization => Data,
-     Post => (if not Error then Data'Initialized);
+.. code:: ada
+
+   procedure Read_T (Data : out T; Error : out Boolean) with
+        Relaxed_Initialization => Data,
+        Post => (if not Error then Data'Initialized);
 
 Additionally, it would allow verifying initialization of data on algorithms
 which defeat the current initialization verification techniques. Indeed, as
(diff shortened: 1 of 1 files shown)


considered/rfc-spark-relaxed-init.rst
```diff
@@ -0,0 +1,520 @@
...
+
+The idea is to have a way to check initialization by proof instead of doing it
+in flow analysis. Using the Relaxed_Initialization aspect allows to define
+precisely which (parts of) an object should be handled by flow analysis or
+proof.
+
+An object has `relaxed initialization` if either:
+
+- it is annotated with the Relaxed_Initialization aspect,
+- it is a formal parameter and it occurs in the Relaxed_Initialization aspect
```
yannickmoy: I would add here the object for a function result, which may be mentioned in the function postcondition.

I would also add X'Old and X'Loop_Entry if X has relaxed initialization.

considered/rfc-spark-relaxed-init.rst
```diff
@@ -0,0 +1,520 @@
...
+  concatenation) on composite types are considered to read the components
+  too.
+
+Initialized attribute
+---------------------
+
+The initialized attribute can be used on any expression with relaxed
+initialization. It is true when all scalar components have been initialized and
+all applicable subtype predicates hold.
+The correct application of this aspect could be checked in the frontend.
```
yannickmoy: "in the frontend" is an implementation rule. We should say here "could be defined as legality rules" which should preferably be checked in the frontend part of GNATprove

considered/rfc-spark-relaxed-init.rst
```diff
@@ -0,0 +1,520 @@
...
+For execution, we could either implement an approximation of this aspect, or
+use Valid_Scalars as a first approximation. For proof, it means adding a flag to
+scalar subcomponents of expressions with relaxed initialization to remember if
+they are initialized or not.
+As it may happen that a scalar is valid even though it has not been initialized,
+so negative occurrences of the Initialized attribute may be interpreted differently
+in proof and for execution. To retain the closest correspondence possible
+between proof and execution, we could avoid assuming that ‘Initialized is
+false in proof when a scalar is not initialized / on out parameters / globals
+of mode Output and rather assume nothing.
```
yannickmoy: I think an example would help here. Just consider:

X : Integer;
pragma Assert (not X'Initialized);

If X'Initialized is interpreted at runtime as X'Valid_Scalars, then the above assertion will fail on most platforms. Indeed, any initial value for X will be a valid integer value. Thus, GNATprove should not be able to prove the assertion.

yannickmoy:
I am of the opinion that 'Initialized should only be allowed inside ghost code (which includes assertions).

considered/rfc-spark-relaxed-init.rst
clairedross: Are those objects and not expressions ?

yannickmoy: according to Ada RM, these are objects.
for 'Old see http://www.ada-auth.org/standards/rm12_w_tc1/html/RM-6-1-1.html#I3370 ("a constant that is implicitly declared")
for 'Result see http://www.ada-auth.org/standards/rm12_w_tc1/html/RM-6-1-1.html#I3372 ("the result object of the function")

ChrisBorden:
Here's a naive comment from an inexperience user (me).  These are relative to Init_By_Proof.  FWIW:
Allows putting Relaxed_Initialization on a type or individual declaration:
+1 - I like that I can use on declaration and don’t have to create subtype, but have the flexibility to add to a type also.

<We have considered using a pragma Annotate (GNATprove, Relaxed_Initialization, V); to mean that a variable V has relaxed initialization, but it was more cumbersome than an aspect, and was a bit complicated (and heavy) when applied to function parameters.>
I agree that aspect is less cumbersome than annotate.  

In general, I like the way this is heading.

jklmnn:
How would arrays be handled with the aspect? Would it be possible to initialize different slices at different times and then prove that the whole array has been initialized?
In regards to arrays I'd also like to be able to prove that
```Ada
for I in Array_Object'Range loop
   Array_Object (I) := Some_Value;
end loop;
```
initializes an array.

yannickmoy:
@jklmnn Array initialization with a simple loop over the array range is now considered as completely initialized after the loop. That was the case for some time when the array is constrained, it's been extended to the unconstrained case not long ago (and it is part of Community 2019). But this is only based on limited heuristics in flow analysis.

With the extension discussed under this PR, you'll be able to very precisely say that cells at such-and-such indexes of the array have been initialized, because you'll rely on the very precise and powerful mechanics of proof. See examples of that with our current prototype in test https://github.com/AdaCore/spark2014/tree/master/testsuite/gnatprove/tests/RC17-014__init_by_proof

PR #23, opened 2019-07-09
[RFC] Proposed expansion of Ada 202X parallel for-loop syntax

sttaft:
Description of proposed expansion of Ada 202X parallel for-loop syntax first into a pragma, and from there into a call on a library routine.


sttaft:
Status as of 6-August-2019: Prototype version of Ada.Parallelism library has been implemented and tested with parallel loop over arrays and vectors, as well as with the N-Queens problem.  Ed Schonberg has not found time to work on adding "Par_Loop" pragma and doing transformation of loop body into a local procedure.

raph-amiard:
In retrospect this seems more like an internal implementation memo, closing this.

PR #27, opened 2019-08-11
[RFC] Simple return by reference syntax

jere-software:
Direct link to text: https://github.com/jeremiahbreeden/ada-spark-rfcs/blob/topic/simple_return_by_ref/considered/rfc-simplified-return-by-reference.rst

### DIFF
--- a/considered/rfc-simplified-return-by-reference.rst
+++ b/considered/rfc-simplified-return-by-reference.rst
@@ -136,6 +136,12 @@ method:
     begin
        return (Element => Source.Something'Access, ...Other stuff);
     end Reference;
+    
+NOTE:  While this proposal focuses on implementing it using the existing
+Implicit_Derefernce types, if it is easier to simply use access types
+under the hood, that is fine.  The intent of this RFC is to reuse existing
+rules, implementations, etc. with minimal implementaiton fuss.
+
 
 Rationale and alternatives
 ==========================
--- a/considered/rfc-simplified-return-by-reference.rst
+++ b/considered/rfc-simplified-return-by-reference.rst
@@ -1,0 +1,165 @@
+- Feature Name: Simplified Return by Reference Syntax
+- Start Date: 2019-08-10
+- RFC PR: (leave this empty)
+- RFC Issue: (leave this empty)
+
+Summary
+=======
+
+Provide a simplified more new user syntax for creating return types that
+return by reference.
+
+Motivation
+==========
+
+The Ada RM defines a "reference type" as "A (view of a) type with a specified 
+Implicit_Dereference aspect" [1].  These types are used to more safely return
+an internal component of an object (such as a container), used for iterators,
+etc.  However, the syntax for such a type is not very user friendly and takes
+a lot of scaffolding that doesn't actually add to reabability.  In fact, it 
+can hamper readability.  New users may find it challenging to create one from
+scratch withough finding and example to mimic, and even then they may not
+fully understand all the implications of the implementation that they end up
+mimicing.
+
+Additionally, using a "reference type" still ends up exposing an access type
+to the client code.  It is safer than a standard access type, but still not
+desirable.
+
+The idea behind this RFC PR is to suggest a simpler, more readable and
+understandable syntax that leverages existing Ada idioms as implementation.
+
+[1]: http://www.ada-auth.org/standards/rm12_w_tc1/html/RM-4-1-5.html#I2450
+
+Guide-level explanation
+=======================
+
+In order to return a component by reference, one should use the syntax:
+
+function Reference(Source : in out Thing) return aliased Some_Type;
+
+This will return a view of an object of Some_Type by reference, presumably from
+Source (though it could come from a package level variable as well if aliased).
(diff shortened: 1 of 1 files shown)


jeremiahbreeden:
I'm not a great write, so I am sure I could be better at laying this out.  Let me know what I need to clarify.  I may ask some questions about how to present it in the RFC though

raph-amiard:
Hi @jeremiahbreeden ! First, thank you very much for your contribution. I agree that making reference types easier to use is a worthy goal.

However:

1. In your RFC you demonstrate how to create a function that returns a reference, but not how to use this result. How would the call site look like ? It seems to me that this is a big hole in your proposal: If you want some kind of short-hand syntax to refer to reference types, it needs to work everywhere.

2. In that light, the use of the "aliased" keyword seems inoportune, since it already has a meaning on a variable declaration, that will be different.

What do you think about this ?

Also I cannot help but think that some version of implicitly instantiated generics might work just fine here, if you had a package `Ada.Containers.Simple_Reference` you could then do

```ada
package R renames Ada.Containers.Simple_Reference;

function Reference(Source : in out Thing) return R (Some_Type).Ref is
   ...
end Reference;
```

This gets rid of the tedious part about reference types - declaring them.

jeremiahbreeden:
First of all, thank you for the comments!

> 
> 
> Hi @jeremiahbreeden ! First, thank you very much for your contribution. I agree that making reference types easier to use is a worthy goal.
> 
> However:
> 
>     1. In your RFC you demonstrate how to create a function that returns a reference, but not how to use this result. How would the call site look like ? It seems to me that this is a big hole in your proposal: If you want some kind of short-hand syntax to refer to reference types, it needs to work everywhere.

Can you please elaborate?  In my proposal, I mentioned it would work exactly as it does for the types that use implicit_dereference (like how it is currently used in Ada containers as one example).  For example, if you had a type that provided the reference function, you could do something like:

Some_Container_Object.Reference(Cursor) := 23;  -- assuming it was an aliased integer being returned

or 

Some_Procedure(Some_Container_Object.Reference(Cursor));  -- assuming an Integer parameter

or if the return type is a tagged record with primitive operations:

Some_Container_Object.Reference(Cursor).Some_Primitive_Op;

And of course, if the Variable_Indexing aspect was defined, then they would just be 
Some_Container_Object(Cursor) := 23; 
Some_Procedure(Some_Container_Object(Cursor));
Some_Container_Object(Cursor).Some_Primitive_Op;

I'm not sure I understand what the hole would be, can you expound?  I don't mind updating the proposal, I just need to understand your question better to address it.

NOTE:  This does not have to be tied to containers or cursors...just handy examples

> 
>     2. In that light, the use of the "aliased" keyword seems inoportune, since it already has a meaning on a variable declaration, that will be different.

I agree, though it isn't the first time we have reused a keyword for something than it is original meaning (see "limited" for example).  I'm not tied to it, I just didn't want to bog the proposal down with a new keyword, cause that tends to degrade discussions off on tangents.  I'm definitely open to using some other keyword or syntax.

> 
> 
> What do you think about this ?
> 
> Also I cannot help but think that some version of implicitly instantiated generics might work just fine here, if you had a package `Ada.Containers.Simple_Reference` you could then do
> 
> ```ada
> package R renames Ada.Containers.Simple_Reference;
> 
> function Reference(Source : in out Thing) return R (Some_Type).Ref is
>    ...
> end Reference;
> ```
> 
> This gets rid of the tedious part about reference types - declaring them.

I don't disagree here.  That sounds like a possible alternative.  I don't know how I would make a package that is used in that way (unless you are also proposing a different syntax here with the R(Some_Type).Ref and are proposing the compiler would provide the package as well)

jeremiahbreeden:
Side note, I am also not opposed to using an aspect instead of "aliased":

function Reference(Container : in out Container_Type; Index : Index_Type) return Some_Type
   with Return_By_Reference => True;

I just wasn't sure this would be as maintainable or readable.  Thoughts?

mosteo:
Another possibility for the return syntax:

`function By_Ref (...) return renames Some_Type;`

In the vein of reusing the existing keywords in the spirit of the proposal.

jeremiahbreeden:
My first reaction feels weird on this, but I am not opposed.  My main goal is to get access types as far from the client of packages as possible, so hopefully the community can come up with a good verbage

> 
> 
> Another possibility for the return syntax:
> 
> `function By_Ref (...) return renames Some_Type;`
> 
> In the vein of reusing the existing keywords in the spirit of the proposal.

sttaft:
For what it is worth, using "return aliased T" feels about right to me, as
we currently use "aliased" to force pass by reference on formal
parameters.  Presumably the rules would mimic closely the current rules on
returning objects of an anonymous access type, or objects with an access
discriminant, to ensure no dangling references.  One issue is whether the
return object is a variable or a constant.  Conceivably you could say it is
a variable by default, and you would have to say "return aliased constant
T" if you wanted it to be treated as a constant.  Not sure that is the
right default, but this could provide a more elegant way to get
user-defined container indexing than the current rather round-about way
using access discriminants.

-Tuck

PR #28, opened 2019-09-11
[RFC] Change Ada.Containers.Count_Type to a subtype of Natural

setton:


raph-amiard:
Hi Nico!

I think your RFC proposes a practical fix to the occurrence of a general issue we have in Ada wrt. operator visibility.

I however wonder if we shouldn't try and address the forest behind the tree here.

Stil in favor of this anyhow. Will put that up for discussion!

raph-amiard:
A problem that I see today with that change, is that it will be very hard to make it affect only one version of Ada, since we can't say:

```ada
% if Ada_Version = 2012 then
   type Count_Type is ...
% else
   type Count_Type is ...
% endif;
```

So while I like that proposal, I think it's unimplementable right now. Curious to know what other people think.

PR #30, opened 2019-10-07
[RFC] Possibly not returning procedure calls

yannickmoy:
Link to full text: https://github.com/yannickmoy/ada-spark-rfcs/blob/possibly_not_returning/considered/possibly_not_returning.rst

### DIFF
--- a/considered/possibly_not_returning.rst
+++ b/considered/possibly_not_returning.rst
@@ -45,7 +45,7 @@ annotation on SPARK procedures that may not return:
      Annotate => (GNATprove, Possibly_Not_Returning);
 
 Such a procedure would be allowed to call a procedure marked ``No_Return``, and
-GNATprove would handle correctly the possibility of non-returning in flow
+GNATprove would handle correctly the possibility of nonreturning in flow
 analysis and in proof.
 
 Motivation
@@ -80,7 +80,7 @@ the tools.
 The new annotation should be described in `SPARK User's Guide appendix
 <http://docs.adacore.com/spark2014-docs/html/ug/en/appendix/additional_annotate_pragmas.html>`_.
 
-A possibly non-returning procedure is defined as a procedure annotated with
+A possibly nonreturning procedure is defined as a procedure annotated with
 ``Possibly_Not_Returning`` as follows:
 
 .. code-block:: ada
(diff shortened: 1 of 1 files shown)


considered/possibly_not_returning.rst
```diff
@@ -0,0 +1,164 @@
...
+Drawbacks
+=========
+
+This feature adds complexity to the tool, but this is deemed as reasonable for
+the expected benefit.
+
+Prior art
+=========
+
+No other proof tools have targetted this issue AFAIK.
```
yakobowski: ACSL has `terminates` clauses for exactly this purpose (section 2.5.4 in version 1.14 of the specification). However, I'm not sure of how this clause is supported by Jessie and the WP plugin. Those clauses are more expressive than the current proposal, but I'm not sure this expressiveness is required.

yannickmoy: indeed thanks! I will mention this clause in the RFC.

sttaft:
Seems like a good solution to the problem.  Could define a new aspect some day if this becomes a widely used capability.
-Tuck

sshyamsunder1988:
Thanks Yannick. I think the proposal looks good. 

However I do have a question regarding the writeup:
1) Only procedures can be "Possibly_Not_Returning" right? However No_Return aspect can be given for a function. 
For example: The scenario I am thinking of is the following:
Function A is No_Return.
Procedure B is "Possibly_Not_Returning" and procedure B calls Function A. 
This is a valid scenario right?

clairedross:
I think No_Return is already only allowed for procedures. As a possible extension, we could consider adding the boolean expression of ACSL if we need something more precise one day. For proof, we would generate an additional postcondition that the boolean no_return is equal to this expression.

yannickmoy:
@sshyamsunder1988 no only a procedure can be marked No_Return

sshyamsunder1988:
> @sshyamsunder1988 no only a procedure can be marked No_Return

I see. I was mistaken then. Thanks for confirming Yannick and Claire. 

Regards
Shreyas

clairedross:
We should take care with dispatching calls. On tagged overidding of procedures, would have to force this pragma to match on the overridden and overriding primitives. How does it work for pragma No-Return currently?

sttaft:
On Wed, Oct 16, 2019 at 5:34 AM Claire Dross <notifications@github.com>
wrote:
>
> We should take care with dispatching calls. On tagged overidding of
procedures, would have to force this pragma to match on the overridden and
overriding primitives. How does it work for pragma No-Return currently?

In RM 6.5.1(6/5-7/5):

6/5   A subprogram shall be nonreturning if it overrides a dispatching
nonreturning subprogram. In addition to the places where Legality Rules
normally apply (see 12.3), this rule applies also in the private part of an
instance of a generic unit.

7/5 If a renaming-as-body completes a nonreturning subprogram declaration,
then the renamed subprogram shall be nonreturning.

---

For possibly-non-returning, it would seem that at least in SPARK the
opposite should apply.  If a subprogram that overrides a dispatching
subprogram is marked as possibly-non-returning, then the subprogram it
overrides should also be marked as possibly-non-returning.  Alternatively,
you could have a separate rule, where a dynamically dispatching call is
presumed to *always* be possibly non returning, unless there is a promise
that it is *not* possibly non returning, purpose with a separate notion of,
say, possibly-non-returning'Class.

For Ada, the goal was to be sure No_Return is not lying, but we don't
*require* that all non-returning subprograms be declared as such.  In
SPARK, I would think that you want to have a No_Return exactly when the
subprogram is non returning.  So the overriding rules in SPARK will
probably need to be stricter, or we'll need some notion of No_Return'Class.

-Tuck

yannickmoy:
thanks @sttaft for pointing out the rule for non-returning subprograms. Actually in SPARK we could allow overridding with both directions of change without impacting soundness, at least in theory. In practice it will be cumbersome to handle mixed cases in derivations, for no visible benefit, so I'd rather impose the same policy on all overriddings of a primitive.

clairedross:
> what about a simpler name for the annotation like May_Not_Return (closer to No_Return) rather than Possibly_Not_Returning?

I like it.

PR #31, opened 2019-10-30
[RFC] Conditional return

jklmnn:
Conditional return statement similar to conditional exits in loops. Full text: https://github.com/jklmnn/ada-spark-rfcs/blob/conditional_return/considered/rfc-conditional-return.rst

### DIFF
--- a/considered/rfc-conditional-return.rst
+++ b/considered/rfc-conditional-return.rst
@@ -1,0 +1,202 @@
+- Feature Name: Conditional return
+- Start Date: 2019-10-10
+- RFC PR: (leave this empty)
+- RFC Issue: (leave this empty)
+
+Summary
+=======
+
+Similar to the `exit when` syntax available in loops a conditional `return when`
+could be useful in procedures.
+
+Motivation
+==========
+
+Linear algorithms or initialization sequences often require subsequent actions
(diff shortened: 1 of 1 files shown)


senier:
Fair enough. On the other hand, loop exit statements could be expressed as `if not Success then exit; end if'`, too. The Ada 83 Rationale states:

> The redundancy provided for conditional exits is itself motivated by readability considerations: loop termination conditions should be marked very conspicuously. Thus, in the recommended paragraphing,

```Ada
exit when CONDITION;
```

> is certainly more conspicuous than the equivalent form in which the exit statement is nested within an if statement:

```Ada
if CONDITION then
  exit;
end if;
```

I wonder why *"loop termination conditions should be marked very conspicuously"* and why the same argument shouldn't also apply to subprogram returns?

mosteo:
There are some advocates of having only one return per subprogram, at the end. If you're going to break this rule, having a 'conspicuous' way of doing it is better. In that case maybe it could be used also for functions:

`return <expr> when <cond>;`

AdaDoom3:
A new RFC encompassing and expanding on this proposal has been added - so I am going to move to close this issue. Any new commentary should be put under this RFC:

https://github.com/AdaCore/ada-spark-rfcs/pull/73

PR #34, opened 2019-11-17
[RFC] use of 'dot' notation on untagged types

sttaft:
RFC on object.op notation being extended to untagged private types, and their completions.  Since this was written, there seems to be a growing consensus to extend this to all untagged types.

Link to RFC: https://github.com/sttaft/ada-spark-rfcs/blob/topic/rfc-prefix-untagged/considered/rfc-prefixed-untagged.rst


setton:
Yes to this! When I first wrote the binding to the `cairo` library, I introduced tagged types where they weren't needed, so that people could use the library in a way that's familiar to them in many other languages... Then I backtracked, thinking this was not morally right (and also potentially introduced an overhead in a performance-sensitive area). So, will look forward to this.

mosteo:
This is beating a dead horse, but it also doesn't help with newcomers that synchronized objects are also 
`object.method` only.

sttaft:
For what it is worth, synchronized interfaces and types derived from them
can use either Object.Op(...) or Op(Object, ...).  In any case, are you
supportive of generalizing Object.Op notation to untagged types?

-Tuck

mosteo:
> are you supportive of generalizing Object.Op notation to untagged types?

Totally, within reasonableness. I get lost with the corner cases discussed in the ARG mailing list.

sttaft:
On Mon, Nov 18, 2019 at 12:42 PM Alejandro R Mosteo <
notifications@github.com> wrote:

> are you supportive of generalizing Object.Op notation to untagged types?
>
> Totally, within reasonableness. I get lost with the corner cases discussed
> in the ARG mailing list.
>
You and me both ... ;-)

-Tuck

jeremiahbreeden:
I'm all for this.  I think it should be extended to all types.  What problems do you expect from non private untagged types?  Can you give some examples with code?  The only issue I have been able to come up with this proposal affects both private and non private, so that distinction doesn't help:

Like if you have a private type that has a full view that is an access type, when you do Object.operation is the operation an operation on the access type or the type the access type points to?  In my mind, I am ok with saying that Object.Op notation for all non tagged types is available except in the cases where the full view is an access type, then package.operation must be used where appropriate and access_value.operation is the normal.

sttaft:
It seems the growing consensus is to allow "object.op" in all cases.
Access types create some ambiguity, but that is true in other circumstances
as well, and disallowing them altogether is awkward without "breaking"
privacy (where rules depend on the nature of the full view of a private
type), which in general would be significantly worse.

-Tuck

jeremiahbreeden:
> 
> It seems the growing consensus is to allow "object.op" in all cases. Access types create some ambiguity, but that is true in other circumstances as well, and disallowing them altogether is awkward without "breaking" privacy (where rules depend on the nature of the full view of a private type), which in general would be significantly worse.
> […](#)
> -Tuck
> On Wed, Nov 27, 2019 at 9:47 PM Jere ***@***.***> wrote: I'm all for this. I think it should be extended to all types. What problems do you expect from non private untagged types? Can you give some examples with code? The only issue I have been able to come up with this proposal affects both private and non private, so that distinction doesn't help: Like if you have a private type that has a full view that is an access type, when you do Object.operation is the operation an operation on the access type or the type the access type points to? In my mind, I am ok with saying that Object.Op notation for all non tagged types is available except in the cases where the full view is an access type, then package.operation must be used where appropriate and access_value.operation is the normal. 

What is the intended method to handle the ambiguity from access types?  If the type is visibly an access type, then the "." operation is a dereference and you must use package.operation notation for ambiguous functions, otherwise it is Object.operation notation maybe?

sttaft:
On Wed, Jan 8, 2020 at 11:27 AM Jere <notifications@github.com> wrote:

> It seems the growing consensus is to allow "object.op" in all cases.
> Access types create some ambiguity, but that is true in other circumstances
> as well, and disallowing them altogether is awkward without "breaking"
> privacy (where rules depend on the nature of the full view of a private
> type), which in general would be significantly worse.
> … <#m_-2447812689092163713_>
> -Tuck
> ...
>
> What is the intended method to handle the ambiguity from access types? If
> the type is visibly an access type, then the "." operation is a dereference
> and you must use package.operation notation for ambiguous functions,
> otherwise it is Object.operation notation maybe?
>
There is actually no Ada rule that says "." after an access value always
implies an implicit dereference.  I would suggest that we shift the rules
on selected components (RM 4.1.3) to allow the identifier for a primitive
subprogram of the type of the prefix after a "." even if the type (say, T)
of the prefix is a named access type, so long as that primitive subprogram
has the first parameter being one of the options now allowed for tagged
types (parameter of type T or access parameter designating T).  This will
create some ambiguity if the access type and its designated type have
primitives with the same identifiers, but that seems relatively uncommon.
We could create a preference, but ambiguity is often better than a
mysterious preference.  There is already a preference for component names
over primitive subprogram names in selected-component notation, and I
presume that would still apply.

My justification for this suggestion is that it is annoying if code that
works fine when you can see only the private type becomes illegal when you
can see the full type.  If the full type for a private type is an access
type, the above suggestion should avoid this annoyance in most cases.

-Tuck

sttaft:
Here is an update that allows object.op() notation on any type.

yannickmoy (approved):
as agreed in LD meeting

rfx77:
I would also allow the dot notation on methods of child packages. there is no real reason why this is not possible since child packages cann access the private part of their parent.

PR #35, opened 2019-11-22
[RFC] Default initialization for out parameters

setton:
This provides a proposal for allowing `out` parameters to have a default initialization, which will be evaluated at the point of subprogram calls. The idea is to eliminate a category of access to uninitialized memory.

### DIFF
--- a/considered/rfc-initialization-for-out-parameters.rst
+++ b/considered/rfc-initialization-for-out-parameters.rst
@@ -1,0 +1,115 @@
+- Feature Name: initialization_for_out_parameters
+- Start Date: 2019-11-22
+- RFC PR:
+- RFC Issue:
+
+Summary
+=======
+
+Default initialization is allowed only for ``in`` parameters. This adds the
+possibility of allowing them for ``out`` parameters,
+
+Motivation
+==========
+
+When writing the body of a subprogram which has one or more ``out``
+parameters, one has to ensure that all these parameters are initialized
+in all the paths that the function could take, otherwise the program
+might access unitialized memory.
+
+This is fragile to maintain: it's easy to miss an initialization in
+the body of complex subpgrograms, and the errors introduced might be
+hard to detect, as unitialized memory might have "by chance" a correct
+value.
+
+Guide-level explanation
+=======================
+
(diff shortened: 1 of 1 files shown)


sttaft:
Looks interesting.  One feature of in-parameter defaults is that you can change them as part of renaming.  If that were allowed for out-parameter defaults, then the out parameter would necessarily be passed "in" as well, initialized to the default value at the call site.  Unfortunately, that would also mean that you could eliminate them as part of a renaming, thereby getting back the possibility of an uninitialized out parameter.  One could restrict renaming to allow you to add or change the default for an out parameter, but not eliminate the default.

Defaults can also be added or removed as part of declaring a generic formal subprogram.  It would be important to decide whether the presence of an out-parameter default would affect matching of actual-to-formal subprograms.

You might also need to think about overriding, and whether an override needs to have an out-parameter default if the original did, and whether this depends on whether the subprogram is a dispatching subprogram.

setton:
Thank you @sttaft for the feedback! I didn't know you could change the defaults of in parameters... For out parameters, it really should be forbidden, since the implementation of the body is strongly dependent on this. Imagine that someone writes a "Frobnize" subprogram with an out parameter "Success", making sure to initialize it to False and only setting it to True in the rare case where Frobnization was successful: we can't have clients change this.

I'll amend the PR to reflect this!

sttaft:
Perhaps you should only allow it on the body when there is a separate spec.  It will be annoying if on renaming you have to always repeat the out-parameter default, with a conforming expression.  The default is really just a convenience for the implementor -- it doesn't do much for the caller.  So perhaps that would argue for allowing it on a subprogram body that has a separate spec, and not bother to mention it in the visible spec at all.

setton:
Yes, that's a good point. Also, the caller doesn't need to know the default value, that's not something useful. Will amend in light of this.

yakobowski:
@setton can you update this proposal according to the last discussion? Also, this idea will be relevant for the WG on initialization.

setton:
Will do.

setton:
Done @yakobowski - proposal updated in light of the discussion with @sttaft.

PR #36, opened 2019-11-25
[RFC] case statement with renamed type values

GyorgyRotter:


yannickmoy:
Hi @GyorgyRotter can you reformat your PR? It does not display well correctly, which makes it hard to follow. See the link at the top for how it is displayed currently. thx

GyorgyRotter:
Hi, please check and read it

yannickmoy:
I fear that others will consider it confusing that you can use "a | a" in a case statement or expression. It seems that the current workaround that you mentioned of using "a .. x", in the quite special case you refer to where x is a renaming of a (and you don't have more renamings), provides what you need, except for the syntax using | instead of ..

GyorgyRotter:
Hi,
I think if the compiler could distinguish between the _a_ and it's renamed type value _x_ we may exclude such cases.
In this case to refer the same type value could be forbidden in this _"a | a"_ form but could be allowed in _"a | x"_ form. 
This could make possible to use multiple renames what the current implementation does not make possible by .. when we are writing this _"when a .. x | a .. y"_ (where _"y"_ is an other rename for _"a"_).
Currently if I understand well for the compiler it is totally equivalent to _"when a .. a | a .. a"_ which is forbidden because of _"when a | a"_ was also forbidden.
With this distinction could produce more readable code in the case of e.g. complex state machines when we could look at the name of the states from different perspectives and we would like to express it in our case statement. I think it would be a nice feature.

yannickmoy:
Hi @GyorgyRotter. I suggest you try to get support for this feature from other Ada developers, so that they weight in this discussion. Otherwise, I fear most people may just push back due to the possibility for undesirable ambiguities that would be possible with this change.

PR #37, opened 2020-01-08
[RFC] Lightweight iterators

reznikmm:
Link to text: https://github.com/reznikmm/ada-spark-rfcs/blob/lightweight-iterators2/considered/rfc-lightweight-iterators.rst

### DIFF
--- a/considered/rfc-lightweight-iterators.rst
+++ b/considered/rfc-lightweight-iterators.rst
@@ -1,0 +1,293 @@
+- Feature Name: lightweight-iterators
+- Start Date: 2020-01-08
+- RFC PR:
+- RFC Issue:
+
+Summary
+=======
+
+We propose new iterator aspects to allow lightweight iterators over a container
+with wellknown interface and abstract API support. Do implement it we sacrifice
+direct access to contained elements, but propose some workarounds.
+
+Motivation
+==========
+
+1. We are looking for a much simpler solution. The inspiration for this proposal is
+   GNAT Iterable_ aspect, but we allow multiple iterators per container.
+
+2. We would like a user iterators for non-tagged container types also.
+
+3. Currently compiler creates and destroyes a cursor object on each iteration, because
+   `Next` is a function. If cursor is something more complex then elementary type
+   it could be costly. Consider a container that keeps records about all its cursors,
+   to avoid memory leeks and keep their states consistent. Then cursor should be
+   a controlled type. We would like to avoid an overhead of useless create/destrory.
+
+4. With current iterators it's very hard to write abstract API (expressed in terms of
+   interfaces) that supports iterators. But if you write it, an implementation
+   will be inefficient (See 3). An example:
+
+.. code-block:: ada
+
+   type Element is limited interface;
+   type Element_Vector is limited interface;
+   --  To express iterable interface over vector you need a cursor
+   --  but you can't provide concrete cursor because you haven't known
+   --  implementation of the vector yet. So Cursor becames an interface.
+   
+   type Cursor is interface;
+   function Has_More (X : Cursor) return Boolean is abstract;
+   
+   function Has_Element (X : Cursor'Class) return Boolean is (X.Has_More);
+   package Vector_Iterator_Interfaces is new
+      Ada.Iterator_Interfaces (Cursor'Class, Has_Element);
+   
+   function Iterate (X : Vector)
+     return Vector_Iterator_Interfaces.Reversible_Iterator'Class is abstract;
+
+With this approach both iterator and cursor should be class-wide types.
+On each iterator compiler will create a new Cursor'Class object, then
+assign it and destroy previous cursor. Such operations can't be
+inlined and optimized. They are useless hard overheads.
+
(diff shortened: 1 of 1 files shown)


jeremiahbreeden:
I like that the suggestion is to decouple iterators from cursors as I feel coupling them together was a huge mistake in Ada's original design.  

I am curious how this works with limited types and large types as it looks like you are simply returning the object by value from a function which is either illegal (limited types) or expensive (large types).

Did I miss something in the text on how that is handled?

Also in your code where you assign iterators to Magic_String, only the first one mentions it is a type and a name.  Is that true for all of the iterators or just the first one?

reznikmm:
@jeremiahbreeden You can [return a limited type] from a function since Ada 2005. We do this _only once_ to construct an iterator variable "in place" before enter the iteration loop. I don't expect an iterator to be a large type. Any way we don't do worse then current implementation does (like Iterate and First functions in the container library).

In Magic_String my idea was that you can use an identifier or a [selected_component] in Iterators aspect. If you use an identifier then it designate both type of the iterator object and iterator to use in the loop statement. If you use a selected_component, then selected_component refers to the type of the iterator object (possible in the other package), but in the loop statement you use just selector_name. So if you have

```ada
limited with Magic_Strings.Word_Iterators;
limited with Magic_Strings.Line_Iterators;

package Magic_Strings is

   type Magic_String is private
     with Iterators =>
       (Each_Character,
        Word_Iterators.Each_Word,
        Line_Iterators.Each_Line);
```

Then last two iterators located in child packages like:

```ada
package Magic_Strings.Word_Iterators is
   type Each_Word is limited private;

   function First (Self : Magic_String) return Each_Word;
   function Has_Element (Self : Each_Word) return Boolean;
   procedure Next (Self : in out Each_Word);

   function Word (Self : Each_Word) return Magic_String;
   function From (Self : Each_Word) return Positive;
   function To (Self : Each_Word) return Positive;

private
```
And usage is

```ada
for Cursor in Text.Each_Word loop
   Ada.Text_IO.Put_Line
     (Cursor.Word.To_String & " starts from " & Cursor.From'Img);
end loop;
```

[return a limited type]: https://www.adaic.org/resources/add_content/standards/05rat/html/Rat-4-5.html
[selected_component]: http://www.ada-auth.org/standards/rm12_w_tc1/html/RM-4-1-3.html#S0098

jeremiahbreeden:
@reznikmm I'm familiar with returns for limited types, but that normally assumes either you are constructing them or you have some means of copying them.  That isn't always the case.  For example, I am given a library that provides a limited type with no copy operation (this is a real world example for me that actually happens very often).  I want to make a collection of them so I can easily parse through 1000 of them to find a particular one (say with a custom Ordered_Map).  With the current iterator scheme, I can do that.  I can make a custom ordered map (searchable by an enumeration or integer type) for my limited types and use a reference_type to access them.  Additionally I can use the "for of" syntax with them.

I'm asking if that is still possible with this method.  It might be, I was just checking because I don't want to lose that functionality.  I'm probably the only person, I make a lot of use out of custom containers for limited types (mainly because that's what all the libraries I am provided with seem to give me).

mosteo:
> I'm probably the only person who cares about this, I make a lot of use out of custom containers for limited types (mainly because that's what all the libraries I am provided with seem to give me).

Unrelated and just curious (sorry for the noise), and I may be missing something obvious, but how are these limited types put into containers in the first place?

jeremiahbreeden:
> 
> 
> > I'm probably the only person who cares about this, I make a lot of use out of custom containers for limited types (mainly because that's what all the libraries I am provided with seem to give me).
> 
> Unrelated and just curious (sorry for the noise), and I may be missing something obvious, but how are these limited types put into containers in the first place?

Most all of the libraries I was given were written in the ada95 days (before support for returning limited types in functions).  A general idiom I have to use for it is:

`
package Lists is new My_Limited_Lists(Some_Limited_Type);
A_List : Lists.List;

...

A_List.Append;  -- notice no parameters here, just an append call to add a default element
Some_Initializing_Procedure(  A_List.Reference(  A_List.Last )  );
`

Since the reference function returns by reference, I can modify it after creating it.  If I add a bunch at a time, I just do a quick "for of" loop to give them all a default values

Not super clean, but limited types are fairly limited (hah!)

clairedross:
I like this idea a lot. I think having several iterators with different values may come in handy, for maps for example. If we want to allow in place modification, we could consider passing a Reference function, returning an access-to-variable type instead of Element. Giving a way to provide a Constant_Reference function may also be interesting for efficiency.

reznikmm:
@jeremiahbreeden I don't see how we can lose the functionality you mentioned. The  returning by value method is used for iterators only. In current edition the RFC hasn't propose a way to access Element from iterator yet. To be discussed. But even with current version you can provide a function, that accepts an iterator value and return user reference type from Ada 2012, so you can access elements by reference.

jeremiahbreeden:
@reznikmm Ok that works then. I just wanted to  make sure.  I wasn't seeing it from the description in the file, so I didn't want it to slip under the radar without at least addressing the topic.  Sorry for the trouble.

PR #38, opened 2020-02-28
[RFC] Fixed array lower bound

QuentinOchem:
This is an initial proposal allowing to fix the lower bound of an array type, in an attempt at fixing some Ada-specific performance footprint.

Link to text: https://github.com/QuentinOchem/ada-spark-rfcs/blob/master/considered/rfc-lower-bound.rst

### DIFF
--- a/considered/rfc-lower-bound.rst
+++ b/considered/rfc-lower-bound.rst
@@ -62,12 +62,6 @@ Array declarations are the same as usual. In particular, when bound are provided
 
   V : A (1 .. 10);
   
-A new form is also provided to omit the lower bound and only provide the upper one:
-
-.. code-block:: ada
-
-  V : A (10);
-  
 With these arrays, slices always slide towards the lower bound. In particular, if you write:
 
 .. code-block:: ada
@@ -109,6 +103,15 @@ Arguably, there’s also a way to achieve this today through a type with discrim
   
 This is however a bit convoluted to write and use.
 
+An alternative would be to use a predicate:
+
+.. code-block:: ada
+
+ type My_String is array (Integer range <>) of Character
+   with Predicate => My_String'First = 0;
+
+This opens other difficulties - a predicate can be an arbitrary condition, this would require the compiler to somehow understand that this specific expression means something. It also means that the predicate has an impact on the type structure, for which there's no provision at this stage. 
+
 Drawbacks
 =========
 
@@ -128,4 +131,12 @@ Nothing specific here.
 Future possibilities
 ====================
 
-Nothing specific here.
+We could introduce ways to ommit the lower bound when declaring an array of a type that has a fixed lower bound. Indeas include:
+
+.. code-block:: ada
+
+  V1 : A (10);
+  V2 : A (<> .. 10);
+  V3 : A (.. 10);
+
+This is more of a "quality of life" / "cosmetic" feature comparted to the initial proposal. If we were going this route, this can be discussed separately.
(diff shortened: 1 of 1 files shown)


considered/rfc-lower-bound.rst
```diff
@@ -0,0 +1,131 @@
...
+
+.. code-block:: ada
+
+  V : A (1 .. 10);
+  
+A new form is also provided to omit the lower bound and only provide the upper one:
+
+.. code-block:: ada
+
+  V : A (10);
```
yakobowski: This syntax is intrinsically ambiguous. `A(n)` could mean either `A (<lower bound> .. N)` (as you probably intended) or "an array of N elements", i.e. `A (<lower bound> .. <lower bound> + N -1)`. I'm not sure it is worth it to introduce another syntax oddity, and I would suggest to drop it altogether.

kanigsson:
There are in fact two ways (I know of) to force a fixed lower bound:
```
   type My_String is array (Integer range <>) of Character;

   type String_Based_Zero (Len : Integer) is record
      Data : My_String (0 .. Len);
   end record;
```

and

```
   type My_String is array (Integer range <>) of Character
   with Predicate => My_String'First = 0;
```

Maybe the compiler could be made smarter to optimize code in both cases. I think your proposed change is very heavy in comparison to the expected performance gain.

jeremiahbreeden:
There are things to consider here though:
>    type My_String is array (Integer range <>) of Character;
> 
>    type String_Based_Zero (Len : Integer) is record
>       Data : My_String (0 .. Len);
>    end record;
> ```
> 
I do this one a lot, but I will say this is one of the few cases where I feel like Ada pushes me into a bad or unfortunate design.  I'm not using a record because my software problem is better solved by a record.  I'm using a record because Ada can't do this any other way.  This adds noise to my designs and makes it harder to understand for the average and novice reader (I've experienced trying to explain what this is first hand).

> ```
>    type My_String is array (Integer range <>) of Character
>    with Predicate => My_String'First = 0;
> ```
> 
I would like this if the compiler could enforce it statically when possible (and ideally without assertions enabled), but this appears to be a dynamic only predicate, which is unfortunate.  But that means Ada would need better compile-time support defined in the language spec.  Sure a compiler vendor could do it themselves, but not being able to rely on the languages rules for stuff like is an awful place to be.  One of the best things C++ did (at least for the embedded landscape) was the constexpr rules.  It gave a specific set of rules for when things "must" be evaluated at compile time by the compiler but the flexibility to let the same code be used dynamically if the inputs were not compile-time.

considered/rfc-lower-bound.rst
jeremiahbreeden: Yeah, this part of the proposal might need to be removed or a different syntax used.

QuentinOchem: Yeah - I agree with @yakobowski and @jeremiahbreeden - I'm not to  happy with that bit either. I'll move this to the "Future possibilities" section.

QuentinOchem:
> There are in fact two ways (I know of) to force a fixed lower bound:
> 
> ```
>    type My_String is array (Integer range <>) of Character;
> 
>    type String_Based_Zero (Len : Integer) is record
>       Data : My_String (0 .. Len);
>    end record;
> ```

This one is actually slightly misleading as Len is really the Last element of the array, not its length. But yes, and documented in the proposal.

> 
> and
> 
> ```
>    type My_String is array (Integer range <>) of Character
>    with Predicate => My_String'First = 0;
> ```
> 
> Maybe the compiler could be made smarter to optimize code in both cases. I think your proposed change is very heavy in comparison to the expected performance gain.

I think this one will open a whole bunch of difficult questions. What if you want more than this condition in the predicate? Do you have another predicate? Overall, I would argue that relying of arbitrary predicate analysis for structural decision is a bit weak.

sttaft:
An existing Ada 202X AI, on hold, has a related proposal:

    http://www.ada-auth.org/cgi-bin/cvsweb.cgi/ai12s/ai12-0246-1.txt?rev=1.3

-Tuck

PR #41, opened 2020-04-04
[RFC] Deferred Partial Instantiation of Generic Specifications

jere-software:
Link to text: https://github.com/jeremiahbreeden/ada-spark-rfcs/blob/topic/deferred_partial_instantiation/considered/rfc-deferred-partial-instantiation.rst

This is to suggest providing a means of creating a simpler specification of a generic by allowing some parameters of the generic to be specified manually while leaving the rest to the client instantiator.  In short it allows for specification: 

```Ada

    generic
       type Type1 is private;
       type Type2 is limited private;
       type Type3(<>);
       with procedure Something(Param1 : Type2; Param2 : Type3);
       with function Image(Value : Type1) return String;
    package My_Package is
       procedure Yay(Value : Type1);  -- calls Image internally
       -- Other Stuff
    end My_Package;

```

to be simplified (notionally) in a later generic:

```Ada

    generic
       type Type1 is private;
       with function Image(Value : Type1) return String;
    package My_Client_Package is
       -- Same public API as before
    end My_Client_Package;

```

using the syntax:

```Ada

    generic
       type Type1 is private;
       with function Image(Value : Type1) return String;
    package My_Client_Package is new My_Package
       (Type1     => Type1,
        Type2     => Integer,
        Type3     => String,
        Something => Something_For_Integer_And_String,
        Image     => Image);

```

without needing to provide scaffolding or package chaining.

### DIFF
--- a/considered/rfc-deferred-partial-instantiation.rst
+++ b/considered/rfc-deferred-partial-instantiation.rst
@@ -52,6 +52,7 @@ Ideally one might want to provide the following API to the userbase:
     end My_Client_Package;
 
 In order to do this the options might include:
+
 1.  Instantiate My_Package publically inside of My_Client_Package.  This
 might be ok in some cases but is awkward as one might have to chain package
 calls (p1.p2.Yay) or perhaps expose Type2, Type3, and P1 to the intended
--- a/considered/rfc-deferred-partial-instantiation.rst
+++ b/considered/rfc-deferred-partial-instantiation.rst
@@ -133,7 +133,113 @@ and can now make the call
 without needing either the internal package instantiation or the 
 API scaffolding.
 
+OPTIONAL (Up for discussion):
+
+Interaction with other generics as a formal package parameter would 
+work seemlessly.  Consider a generic such as
+
+.. code-block:: ada
+
+    generic
+       with package P is new My_Client_Package(<>);
+    package X is
+       -- Stuff
+    end X;
+
+Instances of My_Package that have the same formal values for parameters
+Type2, Type3, and Something can be used as a formal value for package X:
+
+.. code-block:: ada
+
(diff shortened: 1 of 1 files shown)


sttaft:
Nice proposal.  Need to decide how this interacts with formal packages.  In particular, would any instance of the original generic package that matched the partial instantiation be considered an instance of this new generic, and similarly, would any instance of this new generic be considered an instance of the original generic?

jeremiahbreeden:
@sttaft Thanks.  I tried updating the proposal with how to handle it.  I marked it as OPTIONAL.  If it makes the proposal untenable, then it could easily be pulled back out and we could just consider them separate generic package specifications, but I would like to at least explore allowing them to be used interchangeably (with respect to the formal values supplied).  I will say I am getting into a realm where I am not 100% good with the terminology so suggestions on how to fix any of it are appreciated.  Hopefully it at least can be deciphered from the examples if I botched the wording.

sttaft:
Thanks for looking into the implications.  I think you want the matching to
work both ways, pretty much as you have proposed.

Take care,
-Tuck

mosteo:
I have wanted so much this when working on my Rx and Iterators libraries. 

About the extra level of indirection: this already happens sometimes with package renamings, so it's not an entirely new nuisance.

briot:
I had missed that proposal, and saw it mentioned in https://github.com/AdaCore/ada-spark-rfcs/blob/topic/generic_instantiations/meta/rfc-improved-generic-instantiations.md

I also very much like this approach, and indeed missed it when working on the traits-containers library

jeremiahbreeden:
Thanks!  I honestly thought this just got lost in the wind.  Glad it can still poke its head around.  Maybe one day we'll have something similar to it.

PR #51, opened 2020-07-20
[RFC] Distributed Memory

QuentinOchem:
In the context of work with CUDA, we identified limitations in what storage pools were allowing us to do and would like to suggest experimentation around an alternate model to handle distributed memory models.


yannickmoy:
I'd like to discuss what more basic and general enhancements could together bring the requested feature, without resorting to a special-case aspect with special-case handling. For example:

- having a version of storage pools that accepts the address type as parameter
- simplifying the controlled type implementation (at least on some platforms) as suggested separately 
- providing the ability to define type conversions (although that only saves a call to some conversion function in exchange for a type conversion)

Regarding the separation of the object allocation from its type, so that we don't need access types always, it is a topic of general interest that is not limited to this RFC, and should be treated separately IMO.

QuentinOchem:
@yannickmoy it would be interesting to see a full proposal with Storage Pools indeed. One of the reason that pushed us to considering an alternative to object orientation is that - no matter how much you simplify, there's always an additional overhead using OOP (if only because of dispatching). To hold the comparison with alternatives, it would be useful to consider a different approach to pool altogether - this could be an option (then you would replace one special-case aspect, the pool, with another).

I agree with the generalization of the "allocations-with-no-access-type" part of the discussion. If confirm that indeed this is a direction worthy of a specific RFC, I'm happy to draft an initial proposal.

raph-amiard:
I agree with Yannick. The feature looks too specific to me, and we should instead strive to generalize the abstractions that are necessary for your use case. Let's discuss in the specifics.

raph-amiard:
@QuentinOchem This seems to be superseded by the new AI on the storage model, I'll close this.

PR #62, opened 2020-10-07
[RFC] Overriding_Interface_In_Child_Package

avsplunter:
https://github.com/avsplunter/ada-spark-rfcs/blob/master/considered/rfc-interface-child-package.rst


raph-amiard:
Hi @avsplunter,

There seems to be a confusion between the way `interface` is currently used in Ada (to refer to a tagged type interface) and what you mean by it (which seems to be the signature of a package). Could you please clarify that aspect, and maybe add a full example of use of your proposed feature ?

Thanks in advance!

avsplunter:
So, as you probably already noticed, using github is new for me.
I was also overestimating the functionality: replying on an notifications e-mail with the example files will not work.
I no uploaded the example files with a small read.me.

I hope it makes everything a bit more clear.

raph-amiard:
@avsplunter I did not manage to understand the feature yet. However, adding files to the rfc like this is not good, your PR should contain only the rst file, with examples embedded inside of it. Thanks in advance :)

avsplunter:
@raph-amiard I merge the example in the RFC. I also added a short description for the purpose of the request.
I work at Nexperia, so if neccessary we could have a on-line call to explain in person.
PS: sorry about the mess I make with check-in's. It will be better as soon as we start GIT in our own SW development.

PR #73, opened 2021-04-06
[RFC] Conditional when constructs

AdaDoom3:
This RFC is based on the "Conditional return" RFC filed by @jklmnn (AdaCore/ada-spark-rfcs#31) and supersedes it.


Glacia:
Considering that [If expressions](http://www.ada-auth.org/standards/2xrm/html/RM-4-5-7.html) and [raise expressions](http://www.ada-auth.org/standards/2xrm/html/RM-11-3.html) exist i dont think there is any value in this. The only thing missing is goto expressions.

senier:
> Considering that [If expressions](http://www.ada-auth.org/standards/2xrm/html/RM-4-5-7.html) and [raise expressions](http://www.ada-auth.org/standards/2xrm/html/RM-11-3.html) exist i dont think there is any value in this. The only thing missing is goto expressions.

It does have value in environments where exceptions are not available (e.g. a stripped-down runtime or a SPARK program). While if (or case) expressions can solve the issue in some situations, a cascaded initialization procedure (as in the first code example) typically is a sequence of statements.

While I doubt you were serious about `goto`, this could indeed be considered for sake of consistency. I have no strong opinion, though, as I had no need for `goto` in the last 25 years.

Glacia:
Btw, i just realized that this is basically [If modifier](https://ruby-doc.org/core-3.0.0/doc/syntax/control_expressions_rdoc.html#label-Modifier+if+and+unless) from Ruby.

> While if (or case) expressions can solve the issue in some situations, a cascaded initialization procedure (as in the first code example) typically is a sequence of statements.

But in practice those chained if statements tend to be complicated with multiple conditions and with else and elif parts, which make this feature useless in this case. This is basically one liner if statement, that's all it is.

jklmnn:
> Considering that If expressions and raise expressions exist i dont think there is any value in this. The only thing missing is goto expressions.

I absolutely cannot see how an if expression would solve that problem. The proposed solution solves a problem about control flow and not value selection ([An expression is a formula that defines the computation or retrieval of a value.](www.ada-auth.org/standards/2xrm/html/RM-4-4.html#S0132)).
```Ada
if Condition then
   return;
end if;
```
This cannot be replaced with an if expression (if I'm wrong I'd be curious about the counter example).

> But in practice those chained if statements tend to be complicated with multiple conditions and with else and elif parts

Well then the practice I had was different. If you want to call multiple procedures that depend on each others state and break when one of them fails this is actually what you want. There are no else parts be cause we're talking about a return statement. So whatever comes after `end if` is the else part. And if this was so uncommon I wouldn't have taken the effort to create #31.

> This is basically one liner if statement, that's all it is.

It is not. What you call one liner if statement is in reality multiple statements squashed into a single line. It's less readable and also would break style conventions if used. It's the same as suggesting to remove `exit when condition;` in favor of `if condition; then exit; end if;`. Also see https://github.com/AdaCore/ada-spark-rfcs/pull/31#issuecomment-548173651.

senier:
I started trying out that feature and I must say that I love it. Especially termination conditions become much more concise as you first state what is to happen (`return`) and then under which condition (`if Value = 0`). With proper indentation this is much more readable IMHO (and you safe a lot of vertical space).

raph-amiard (approved):

Issue #26, opened 2019-08-07
[RFC] Compile time string formating


Fabien-Chouteau:
So I was thinking about string formatting in Ada and I want to discuss the concept below. I don't know if something similar was already proposed.

The idea would be to add something that is essentially syntactic sugar.

**Syntax is discussed at the end.**

This: 
```ada
function Plop (Name, Profession : String) return String
is ("\s is \s" with (Name, Profession));
```
Whould be equivalent to this:
```ada
function Plop (Name, Profession : String) return String
is (Name & " is " & Profession);
```

This would be a compile time feature, so only applicable to string literate.
This is **not** allowed:
```ada
function Plop (Fmt, Name, Profession : String) return String
is (Fmt with (Name, Profession));
```

### Bonus 1 : escape characters

We could add escape characters: `\n` for `ASCII.LF`, `\t` for `ASCII.HT`, etc.
```ada
function Plop (Name, Profession : String) return String
is ("\t\s is \s\n" with (Name, Profession));
```
vs
```ada
function Plop (Name, Profession : String) return String
is (ASCII.HT & Name & " is " & Profession & ASCII.LF);
```

### Bonus 2 : Automatic images

`\i` could mean to automatically apply attribute `'Img`:

```ada
function Plop (Name, Profession : String; Since : Positive) return String
is ("\t\s is \s since \i\n" with (Name, Profession, Since));
```
vs
```ada
function Plop (Name, Profession : String; Since : Positive) return String
is (ASCII.HT & Name & " is " & Profession & " since " & Since'Img & ASCII.LF);
```
### Bonus 3 : Inserts length

We could specify the desired length for the inserted text:

```ada
function Plop (Name, Profession : String; Since : Positive) return String
is ("\t\20s is \s since \i\n" with (Name, Profession, Since));
```
(I don't know an easy way to do that with current language features)

### Syntax

There is two syntax aspects to this feature:

#### Format operator
This is the relation between the format string `"\s is \s"` and the strings to be inserted `(Name. Profession)`.

I see multiple ways to do that:
 - An existing keyword: `"\s is \s" with (Name, Profession)`
 - A new operator: `"\s is \s" % (Name, Profession)`
 - Attribute: `"\s is \s"'Format (Name, Profession)` (the combination of double and single quote is quite ugly to be honest)
 - A la subprogram call: `"\s is \s" (Name, Profession)`

#### Token

[...]

PR #77, opened 2021-06-01
[RFC] String Interpolation

sttaft:
This is the first draft of an RFC for the string interpolation feature.


glacambre:
Link to the rendered version: https://github.com/sttaft/ada-spark-rfcs/blob/topic/rfc-string-interpolation/considered/rfc-string-interpolation.rst

sttaft:
Rendered version also here: 
https://github.com/AdaCore/ada-spark-rfcs/blob/bfe06abe9f04ea26a2351cf4a7f6bf091b13d193/considered/rfc-string-interpolation.rst

Fabien-Chouteau:
Thanks @sttaft ,

My two cents on this proposal.

# 1. Referencing variable names
I am not a huge fan of referencing variable names in the template. I think especially with Ada and its most common programming style, names tend to be very long and will result in a template string that is not readable.
If I take an example from the original issue:
```ada
Put_Line ($"(Name) is a (Profession)"$)
```
I think it's going to be pretty rare to have such a clean template. In practice you will have names that are part of a record, or something else more complex. So in my opinion we will most like see patterns that will look like this:
```ada
Put_Line ($"$(My_Object.Name) is a $(Professions_Img_Array (My_Object.Profession_Index))"$);
```
Of course there will be the options to declare renames for strings:
```ada
declare
   Name : String renames My_Object.Name;
   Profession : String renames Professions_Img_Array (My_Object.Profession_Index
begin
   Put_Line ($"$(Name) is a $(Profession))"$);
end;
```
But in the end I don't think this brings as much readability improvement as we could get.
Of course I would like to have other opinions on this.

I see two other way around, and I am introducing a placeholder "syntax" `$"  "$ (...)` just for the examples:
 - ### a. Positional:
   ```ada
   Put_Line ($"$(0) is $(1)"'$ --  This is a lot of `$` I must say
             (My_Object.Name, 
              Professions_Img_Array (My_Object.Profession_Index));
   ```

 - ### b. Another variation on naming (I saw this in Python at least)
   ```ada
   Put_Line ($"$(Name) is $(Profession)"'$
             (Name       => My_Object.Name, 
              Profession => Professions_Img_Array (My_Object.Profession_Index));
   ```
# 2. Expression in templates:
```ada
Put_Line ($"Plop is $(X + Y)"$);
```
I the same vein as my comment above, I think this is likely to generate templates that will be hard to read. And I am not an implementer, but I am guessing that this will introduce complexity that can be avoided in my opinion.

# 3. Formatting options
In the RFC ` $(X+Y, Width => 13)` is given as an example. I am personally a fan of printf's `%2.3f`, `%5x`, `%04d`, etc.
If we can make something as compact as that it would be great.

Looking in Python's direction, it could be something like:
```ada
Put_Line($"Plop is $(0.2f:Angle)"$)
```

# 3. Embedded
As most of you know my focus is on embedded and in particular bare-metal and small device. So I am always worried about how a given feature can be implemented and used in this kind of situation. I don't know how compatible with embedded the current proposal is, but I just want to state that it is important to keep that in mind.

mgrojo:
I hope editors start to highlight the interpolated variables different to the characters around, otherwise this would be less readable instead of more.

I suppose escape characters (i.e. `"\t\n"`) will not work in traditional string literals, for compatibility. But then people new to the language will start to complain about being unable to write `"\tHello, World!\n"`. Is the introduction of this C-ism really needed?, wouldn't using `ASCII` entities be more respective with current Ada: `use ASCII; $"$(HT)Hello, World!$(LF)"$`?

> This is syntactic sugar, so there is no magic here. It is going to be equivalent to a string concatenation and calls on 'Image, and the programmer will need to keep that in mind.

Does that mean that interpolated positive integers will have a leading space? I would understand it for the shake of consistency, but that would shock many.

sttaft:
On Sat, Jun 19, 2021 at 8:16 AM Manuel ***@***.***> wrote:

> ...
>
> This is syntactic sugar, so there is no magic here. It is going to be
> equivalent to a string concatenation and calls on 'Image, and the
> programmer will need to keep that in mind.
>
> Does that mean that interpolated positive integers will have a leading
> space? I would understand it for the shake of consistency, but that would
> shock many.
>

No, there would be a "Trim" applied to the result of 'Image so no extra
spaces.  The RFC should have made that clear.

-Tuck

considered/rfc-string-interpolation.rst
```diff
@@ -0,0 +1,192 @@
...
+adding complexity to string literals, and 
+
+Prior art
+=========
+
+String interpolation has begun to show up in many languages.  Python has a number
+of string literal syntaxes, chosen by a prefix letter, but our sense is that
+the string interpolation syntax has emerged as the favorite.  We do not want
+to have lots of different syntaxes, so we have included the escape mechanism
+as part of both of the new string literal syntaxes.  We have chosen '\' as the
```
mgrojo, line 180: This is rendered as  `''`. It probably needs to be quoted using backticks like this: `\`

sttaft:
Thanks.  I fixed it in both places, and added a mention of trimming leading
and trailing white space.

-Tuck

raph-amiard:
* Overall, looking really good!
* ‘Image doesn’t help when you want to interpolate say an unbounded string into a string, because it’ll typically have quotes around it, but you don’t want that. How do we accommodate this common use case?
* One drawback about multiline strings is that they’ll be the first multi-line token in the Ada language. Single line tokens only makes it very easy to tokenize/highlight Ada so far.
* Not a fan of the terminating $

Fabien-Chouteau:
@zertovitch I was just thinking about this, maybe it would be interesting to prototype this string interpolation in [HAC](https://github.com/zertovitch/hac).

zertovitch:
Why not? A few hints on how to do it: there is an internal type called 
String_Literals.

For adding the string interpolation feature, look at the lines 163+ 
(currently) in src/hac_sys-parser-expressions.adb .

You can add a boolean mode switch, say `Interpolated` activated by the 
front '$', then parse the literal string in the special mode if 
Interpolated = True.


             procedure Primary (FSys_Prim : Symset; X : out Exact_Typ) 
is    --  RM 4.4 (7)
               F   : Opcode;
               Ident_Index : Integer;
             begin
               X := Type_Undefined;
               Test (CD, Primary_Begin_Symbol + StrCon, FSys_Prim, 
err_primary_unexpected_symbol);
               case CD.Sy is
       >>>       when StrCon =>
                   X.TYP := String_Literals;
                   Emit_1 (CD, k_Push_Discrete_Literal, Operand_2_Type 
(CD.SLeng));  --  String Literal Length
                   Emit_1 (CD, k_Push_Discrete_Literal, Operand_2_Type 
(CD.INum));   --  Index To String IdTab
                   InSymbol (CD);
                 when IDent =>

dsibai:
As discussed with Yannick/Tucker, the 'double $' syntax is extremely difficult to read.

I'd rather either have a letter indicating a formatting string such as F :
`F"This is $X and $(X+Y)"`

or a whole word :
`Format"This is $X and $(X+Y)"`

The main problem with the word approach is that it looks like a mistyped function call, so a special, compiler-supported function  such as
`Format("This is $X and $(X+Y)")`
might be more readable.

I'll also add that bash/python/rust/etc reflexes make we want to write:
`F"This is ${X} and ${X+Y}"` or `Format("This is ${X} and ${X+Y}")`

yannickmoy:
> As discussed with Yannick/Tucker, the 'double $' syntax is extremely difficult to read.

agreed

> I'd rather either have a letter indicating a formatting string such as F : `F"This is $X and $(X+Y)"`

let's see what others think?

> or a whole word : `Format"This is $X and $(X+Y)"`

I also like that, as it is clearer that this starts a format string.

> The main problem with the word approach is that it looks like a mistyped function call, so a special, compiler-supported function such as `Format("This is $X and $(X+Y)")` might be more readable.

That would conflict with a user function called Format taking a string and returning a string, so I don't think we can do this.

> I'll also add that bash/python/rust/etc reflexes make we want to write: `F"This is ${X} and ${X+Y}"` or `Format("This is ${X} and ${X+Y}")`

I'd rather avoid useless curly braces, unless there is both an advantage in readability and uniformity.

glacambre:
>I'd rather either have a letter indicating a formatting string such as F : F"This is $X and $(X+Y)"

I'm not a fan of format specifiers. What do you think of using backticks? e.g.
```
Put_Line (`X = $(X)`);
```

This way interpolated strings are as light as regular strings in terms of syntax but still distinguishable from regular strings.

raph-amiard:
I'm not a fan of backticks because they introduce a new token in the lexer, and probably new escaping rules in string. I prefer prefixed strings simply because they're simpler to implement, and also pretty familiar to people because of Python. They're also a more generic syntax extension, that you might be able to reuse if you need another type of literals someday.

kevlar700:
I can't say that I have analysed this discussion in great detail but I shall throw in a couple of cents from my experience with Go and Dart.

Dart has string interpolation and Go does not.

Backticks allow multi line strings in Go, perhaps a multiline feature, could allow the nicer interpolation that Dart has with longer names?

Go has a printf but it's general use is type aware and so you can do this but I often miss spaces out.

fmt.Print("Hello", worldVar, "oops worldVar has no leading space")

dsibai:
I found this interesting comment by Claire Dross:
https://github.com/AdaCore/ada-spark-rfcs/issues/26#issuecomment-519007316
She suggests having an attribute 'Format.

So we could have an interpolation function as an attribute:
String'Format("$(X+Y) = $(X) + $(Y)")

- We'd have a readable word
- It would look like a function and wouldn't look mistyped like Format"..."
- More importantly, the function would be allowed to be magical, since it's an attribute
- There couldn't be any confusion with a regular Format(String)-> String function.

mhatzl:
> So we could have an interpolation function as an attribute: `String'Format("$(X+Y) = $(X) + $(Y)")`

I like this syntax mentioned by @dsibai 

As an addition, the formatting options mentioned by @sttaft could optionally be used like `String'Format("$(X+Y, "3.4E2") = $(X) + $(Y)")`

> Seems like an interesting approach.  It is somewhat redundant to specify
> the type (with the "f") in this case, and using letters for specifying the
> radix seems a bit odd given that Ada currently never does that, but some
> kind of Ada-oriented pattern such as "3.4E2" for specifying the format for
> a floating-point number, and "16#5#" for specifying a hex integer might be
> pretty intuitive.  "*" might be used for cases where the programmer doesn't
> care, such as "16#*#" for hex, but no specified number of digits.  We
> probably still need to think about whether we want to pass this to the
> 'Image function, or to do it all with postprocessing (which is clearly
> harder for some of these patterns).

Having the option to set formatting options by passing a variable instead of a literal might also be useful.
Like `String'Format("$(X+Y, myFormat) = $(X) + $(Y)")`. However, I am not sure how wrong formatting options should be handled in this case.

pyjarrett:
A major strength of Ada is the ability to intuit what code is doing due to minimum symbology and that many of the things you come across can imminently be thrown into a search engine with good results.  Adding a dollar-sign introduces additional symbology while reducing readability, and complicating searching for this syntax.  From one of these, I also can't tell the type of the formatted string.  In this way, I don't think it agrees with the Ada principles of readability or strong typing.

Both [Rust](https://doc.rust-lang.org/rust-by-example/hello/print.html) and [C++](https://en.cppreference.com/w/cpp/utility/format/format) allow a format function, with contained elements in braces, followed by a variable number of parameters in the call.  Indexing parameters by integers tends to be very error prone.

 In usual fashion, I would want to describe exactly what I'm doing, with double braces being a literal brace:
 ```
 Put_Line (String'Format ("{X + Y} = {Y} + {Y}"));
 ```

For a multi-line string, it's formatting multiple lines, it would be nice to just have `String'Format_Multiline`, but there probably does need to be a new type of string literal to cross lines.  Python uses triple quotes.  The `$" ... "` syntax might be useful here.

Format specifiers are convenient and terse, but can be painful to remember to do correctly and unintuitive.  Combining this with named keys like in Rust or [python f-strings](https://www.python.org/dev/peps/pep-0498/) makes this read like what it does without any googling:

```
-- Using "when" to indicate named parameters with attached aspect-like
Put_Line (String'Format("{Sum} = {X} + {Y}",
    when Sum => X + Y with Justify => Right, Fill => '0', Width => 5, Precision => 3,
    when X => X with Justify => Right, Fill => ' ', Width => 4,
    when Y => Y with Justify => Left,
    when others => <> -- use interpolation for anything left over, only needed if any renames exist
   ));
```

```
-- Renames to indicate named parameters with attached aspect-like
-- Parameters with aggregate-like to remove ambiguities due to commas
Put_Line (String'Format("{Sum} = {X} + {Y}",
    renames
    Sum => X + Y with (Justify => Right, Fill => '0', Width => 5, Precision => 3),
    X => X with (Justify => Right, Fill => ' ', Width => 4),
    Y => Y with (Justify => Left),
    others => <> -- use interpolation for anything left over, only needed if any renames exist
   ));
```

Obviously neither is a final syntax, just an idea.

godunko:
It is expected that "string" is a literal, and had one of standard string types which is not a case for i18n applications. In such case string literal need to be translated into another native language and it is not reasonable to use predefined string (array) types here (it is at least not safe in terms of stack overflow).

Another unresolved thing is use of some kind of formatter, which is able to use locale settings to generate output expected by the user (expected 'dot' character in numbers, thousands separator, measure units, etc.).

yannickmoy:
> It is expected that "string" is a literal, and had one of standard string types which is not a case for i18n applications. In such case string literal need to be translated into another native language and it is not reasonable to use predefined string (array) types here (it is at least not safe in terms of stack overflow).

Then what are you proposing for that case?

> Another unresolved thing is use of some kind of formatter, which is able to use locale settings to generate output expected by the user (expected 'dot' character in numbers, thousands separator, measure units, etc.).

This is related to localization, for which I think we need a library approach rather than this simpler format-string feature discussed on this PR.

onox:
> This is related to localization, for which I think we need a library approach rather than this simpler format-string feature discussed on this PR.

I agree that a library is more appropriate and flexible to deal with translations and language-related formatting of numbers, plural rules for numbers, currencies, units, date/times, relative times, etc.

reznikmm:
A few comments:
* **Issue:** meaning of `\n` is unclear
   What character(s) is [new line](https://en.wikipedia.org/wiki/Newline) represented by? ARM allows a lot of characters as _new line_ separator:
    > * 2.1 (16/3) The character pair CARRIAGE RETURN/LINE FEED (code points 16#0D# 16#0A#) signifies a single end of line (see 2.2); every other occurrence of a format_effector other than the character whose code point position is 16#09# (CHARACTER TABULATION) also signifies a single end of line.
    > * 2.1 (13/3) format_effector The characters whose code points are 16#09# (CHARACTER TABULATION), 16#0A# (LINE FEED), 16#0B# (LINE TABULATION), 16#0C# (FORM FEED), 16#0D# (CARRIAGE RETURN), 16#85# (NEXT LINE), and the characters in categories separator_line and separator_paragraph.

   If we hard-code `\n` as `LF` it will make no sense on Windows (and, perhaps, on other platforms). If we hard-code it as `CR+LF` it will make no sense on Unix (and, perhaps, on other platforms). There is **no** _new line_ character common to all platforms. Let this meaning to be implementation defined is even worse, because every usage of `\n` makes the program non-portable.
   **Solution:** Exclude `\n` from this RFC at all.

 * **Issue:** New line representation in the multi-line string literal. The same as for `\n`.
   **Solution:** Introduce a new type (or interface?) String_Vector with a `Join` method. The `Join` should accept desired separator as a parameter and returns a concatenation of all elements with the separator included between them. Let the multi-line string literal return String_Vector. Each element in the vector is a separate line with no _new line_ character at the end. See for [Virtual_String_Vector](https://github.com/AdaCore/VSS/blob/master/source/text/vss-string_vectors.ads#L87) as an example.
   ```ada
   function Join_Lines
     (Self           : Virtual_String_Vector'Class;
      Terminator     : VSS.Strings.Line_Terminator;
      Terminate_Last : Boolean := True)
      return VSS.Strings.Virtual_String;
   --  Join all string vector's strings with each element separated by given
   --  Terminator. When Terminate_Last is True line terminator is added after
   --  last line.
   ```

 * **Issue:** Why `\"` if we already have `""`? This is just an extra complexity in the language.

 * (Minor issue): Use `$$` instead of `\$` as more inline with `""` escape sequence.
   If we drop `\$`, `\"` and `\n` then we don't need `\` escape character any more. This way Ada source code generation becomes simpler, because you need to escape only two characters (`$`, `"`) instead of three.

Fabien-Chouteau:
>     * Let the multi-line string literal return String_Vector. Each element in the vector is a separate line with no _new line_ character at the end.

This is interesting. What would the String_Vector type look like? If it's based on the Ada.Containers it's a no-go for me.

godunko:
> 
> Then what are you proposing for that case?
> 
Interesting solution is general "syntax sugar" for replacable library implementation. For example, user defined string types may provide definition for 'Format attribute.

onox:
Does `{"..."}` provide some safety benefit over `F"..."` or is it just a preference? I think `F"..."` is preferred because it is similar to how other languages do it as @pyjarrett said. (New) Ada users who would see the syntax for the first time can probably guess what it does if they have experience with one of these languages.

yannickmoy:
> Does `{"..."}` provide some safety benefit over `F"..."` or is it just a preference? 

It's just a syntax discussion, no change in semantics.

> I think `F"..."` is preferred because it is similar to how other languages do it as @pyjarrett said. (New) Ada users who would see the syntax for the first time can probably guess what it does if they have experience with one of these languages.

That's also my preference. Others prefer `{"..."}` because Ada does not otherwise use letters as syntactic elements, but I think that's the same for other languages using these F-strings.

sttaft:
By the way, it seems a bit inconsistent to avoid use of "\\" even though it is widely used in other languages, and then say that we should *not* use {" ... "} because it is not used widely.  I believe we should focus on readability and usability for Ada, and make choices that are sensitive to both Ada's tradition (e.g. mirrored syntaxes for bracketing syntax) and to conventions adopted widely (e.g. "\\"), with readability being a very high priority.

One problem with doubling is that you always have to know exactly which characters are and are not to be doubled, whereas with "\\" you can presumably use it with characters that you are not sure about.  So for example, in these new strings, would you have to double "}"?  It is not really necessary.  And what about single apostrophe (')?

sttaft:
My own preference is to strip both leading and trailing whitespace, as it seems the simplest and most uniform rule, allowing the implementor of 'Image to use extra whitespace for standalone situations where some kind of alignment might be appropriate.

I think it might be a bigger surprise for a user if they write:

     {"The input is {X} and the result is {F(X)}."}

and they end up with extra white space before or after the interpolation, when clearly the user did not expect that since they allowed for exactly one space on either side.

One simple way, if you really want to see all of the whitespace that 'Image produces, might be to write:

    {"The image of the input is {X'Image} and the image of the result is {F(X)'Image}"}

But normally I would say you are more interested in seeing the *value* of X or F(X), not its verbatim *image* which might have additional whitespace to provide some kind of formatting in a standalone usage.

The other option, as we have discussed, would be to allow for a second parameter inside the { ... } to provide control over how the 'Image is transformed during interpolation:

    {"The image of the input is '{X, Verbatim}' and the image of the result is '{F(X), Verbatim}'"}

onox:
> e.g. use concatenation between F-strings to indicate newline: `F"hello" & F"world"` instead of `F"hello\nworld"`.

I think it would be surprising to the user if implicit newlines appear in the output because of concatenation. Also how would you avoid an undesired newline if you need to split some F-strings into multiple parts to avoid long lines (> 79 chars) because of style checks (`-gnatyM`)?

yannickmoy:
Hi Tuck, currently it reads "trimmed of a leading space if the second character is a digit", I assume that's just a leftover from the previous version?

Also, you allowed the use of "" for double-quote character in format strings, and we discussed that it was also a leftover.

yannickmoy:
@sttaft you mention in the first paragraph the use of backslash to enter unicode characters, is that intentional? It's not described in the rest of the RFC, and we did not discuss it.

yannickmoy:
You also mention \n as the character used for newline in multi-line strings, but shouldn't it depend on the platform, like the characters inserted by `New_Line`, so that it would be \r\n on Windows?

yannickmoy:
small fix: the rule for `interpolated_string_literal` should have double-quotes at start and end, not only curly braces.

yannickmoy:
You say: "An escaped_character represents the given graphic_character" which seems to indicate that `\k` would be interpreted as just the character `k`. Shouldn't we ask for an error to be issued in such a case, and have explicitly a rule that `\\` and `\"` denote respectively the backslash and double quote?

considered/rfc-string-interpolation.rst
```diff
@@ -0,0 +1,243 @@
...
+Summary
+=======
+
+Provide a new string literal syntax which supports the use 
+of "string interpolation," where the names of variables or
+expressions can be used directly within the string literal, such that
+the value of the variable or the expression is "interpolated" directly
+into the value of the enclosing string upon use at run-time.  In addition,
+an escape character ('\\') is provided for inserting certain standard control
+characters (such as newline) or unicode characters within
```
yakobowski, line 15: This is the unicode reference @yannickmoy is talking about.

sttaft: I moved this discussion to "future possibilities".

raph-amiard (approved):

PR #82, opened 2021-09-20
Improve instructions for creating a RFC

yakobowski:


yakobowski:
Thank you @sttaft. I did address the typos for the moment

> You don't explain how to revise an RFC using your new "simplified" approach.

That's correct. I too think this should be addressed, but this will require a brand new "technical" part. I will see where/how I can add this.

yannickmoy (approved):

PR #83, opened 2021-10-07
Explain in more details where RFC live and the process

yannickmoy:


README.md
```diff
@@ -34,6 +34,17 @@ users can give and get feedback on RFCs. It will also make it possible for
...
 
+You can have a look at:
+
+- [all RFC (both open and closed)](https://github.com/AdaCore/ada-spark-rfcs/pulls?q=is%3Apr+)
+- [open RFC](https://github.com/AdaCore/ada-spark-rfcs/pulls)
+- [RFC being considered for prototyping](https://github.com/AdaCore/ada-spark-rfcs/tree/master/considered)
+- [RFC which have been prototyped](https://github.com/AdaCore/ada-spark-rfcs/tree/master/prototyped)
+
+If you're new to GitHub, you may want to have a look at [GitHub
+Guides](https://guides.github.com/), in particular to learn how to fork the
```
yakobowski, line 45: I would remove this. With the new workflow, the project is implicitly merged, users no longer need to do that. (And those familiar with Github workflow won't need to refresh their memory anyway.)

PR #94, opened 2022-11-23
[RFC] Max_Size aspect for tagged types

raph-amiard:
Link to text: https://github.com/AdaCore/ada-spark-rfcs/blob/topic/max_size/considered/rfc-class-size.md

### DIFF
--- a/considered/max_size.md
+++ b/considered/max_size.md
@@ -1,0 +1,99 @@
+# Max_Size aspect for tagged types
+
+## Summary
+
+The idea is to be able to annotate a tagged type declaration with a `Max_Size`
+aspect like so:
+
+```ada
+type Foo is tagged abstract null record 
+with Max_Size => 16; --  I imagine the size would be in bytes
+```
+
+Then, any time a descendent would be declared, the compiler would verify that
+it satisfies the size constraints
+
+```ada
+type Bar is new Foo with record
+   S : String (1 .. 128);
+end record; --  ERROR: Record doesn't fit in 16 bytes
+```
+
+The usefulness of this is that the classwide type for `Foo` would be of a
+definite and known size, which would allow the users to use classwide instances
+as definite sized objects:
+
+```ada
+Inst : Foo'Class;
+type Foo_Array is array (Positive range <>) of Foo'Class;
+```
+
+NOTE: Manually computing the size is tedious, so the idea would be to leverage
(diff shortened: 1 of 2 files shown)


considered/max_size.md
```diff
@@ -0,0 +1,99 @@
+# Max_Size aspect for tagged types
+
+## Summary
+
+The idea is to be able to annotate a tagged type declaration with a `Max_Size`
+aspect like so:
+
+```ada
+type Foo is tagged abstract null record 
+with Max_Size => 16; --  I imagine the size would be in bytes
```
Fabien-Chouteau, line 10: As far as I know, all Size aspects in Ada an in bits and not bytes, so this one should be the same I guess.

raph-amiard: Good point!

mosteo:
I can see myself easily abusing this one to have definite classwide objects, just like tagged is sometimes used just to get dot notation. Maybe because I do think that Ada would greatly benefit from having a native way to use indefinites as definites, beyond the clunkly `Indefinite_Holders`. 

I guess my kind-of-a-point is that this proposal seems very specific but the interest on what it does may be of a wider scope.

considered/max_size.md
```diff
@@ -0,0 +1,99 @@
...
+### Syntax
+
+No custom syntax
+
+### Static legality rules
+
+* `Max_Size` can be specified only on tagged type definitions. Any use on other
+  entities will raise an error.
+
+* `Max_Size` can only be specified on the rootmost type of a tagged type
```
jklmnn, line 75: Maybe it's possible to allow this aspect on derived types under the restriction that it only can reduce the size? It's just an idea, I'm not sure how useful this would be in practice.

raph-amiard:
@mosteo 

>  I can see myself easily abusing this one to have definite classwide objects, just like tagged is sometimes used just to get dot notation. Maybe because I do think that Ada would greatly benefit from having a native way to use indefinites as definites, beyond the clunkly Indefinite_Holders.

> I guess my kind-of-a-point is that this proposal seems very specific but the interest on what it does may be of a wider scope.

You're correct. This proposal is intended to be one of a bunch that will allow constrained versions of types that were necessarily unconstrained so far (see https://github.com/AdaCore/ada-spark-rfcs/pull/97 for another example)

considered/max_size.md
raph-amiard: I'll mention this in "future possibilities"

PR #100, opened 2023-02-10
Create RFC for Embed attribute

dkm:
First draft for the 'Embed attribute (inspired by C's #embed, Rust's include_bytes!() and probably others).


dkm:
[Rendered document](https://github.com/dkm/ada-spark-rfcs/blob/marc/rfc_embed/considered/rfc-embed-binary-resources.rst)

dkm:
> As mentioned in another email chain about this, I would recommend avoiding a construct that looks executable/evaluable such as an attribute. The implication would be that each time it is encountered in the flow of execution, it is re-evaluated, implying reading a file at run time rather than compile time, which is clearly not the intent.
> 
> An alternative approach is to use an aspect on the declaration, with semantics approximating "Import" though without implying the object is allocated and initialized outside of Ada. Rather, we are saying it is a regular Ada object, but its initialization expression is defined in an external file. For example:
> 
> ` Some_Other_Data : Some_Record with External_Initialization => "/some/data.raw";`

This was the main subject when this RFC was discussed... IIRC, the main objection to the aspect is that you can only use it on the declaration.

@jklmnn proposed something like (extending the `Import` aspect/pragma) : 

```Ada
Some_File_Data : Byte_Array with
   Import,
   Convention => File,
   External_Name => "/some/file/data.raw";

pragma Import (Some_File_Data, File, "/some/file/data.raw");
```

raph-amiard:
Yes there was indeed (very relative) agreement that the aspect version was a bit more flexible in how it could be used, for no visible drawback.

@sttaft While I see your point for me it's not generally clear anyway what "evaluation of an expression" entails in modern languages, so this is not a compoundly very big problem IMO.

OTOH I'm not sure being able to use this directly in expressions yields an expressive power that is very useful in most cases anyway.

I was not fond of re-using the `Import` aspect though, so if we go this way, I do indeed prefer having a specific aspect for this purpose.

raph-amiard:
> Did you mean the "attribute version" above?

Yes

> I don't understand this comment. Could you clarify?

What I mean is that we could, I think, define the evaluation of embed in such a way that the read is done only once, and every subsequent use is only a syntactic "inlining" of the content.

Furthermore, the version with an aspect does suffer from similar problems, if we don't restrict its use to the top level, or at least to constants:

```ada
procedure Foo (J : Integer) is
   Some_Other_Data : Some_Array with External_Initialization => "/some/data.raw";
begin
   Some_Other_Data (J) := 15;
   Do_Something_With (Some_Other_Data);
end Foo;

procedure Bar is
begin
   for J in 1 .. 100 loop
      Foo (J);
   end loop;
end Bar;
```

> That was my sense. Personally I saw many complexities in trying to give the attribute version a good definition in Ada such that it was certain to be performed on the host machine at compile time, and I am sure Steve Baird could come up with an order of magnitude more complexities than me.

We discussed it again, taking your feedback into account, yesterday, and discussing other potential topics, such as mutability, constness, where the data is located (stack or not), validity of the result, and the version I was most satisfied with was something akin to:

* We use an aspect like `External_Initialization` in your example above
* The use of object declarations with such an aspect is restricted to library level packages (not a big problem for the kind of uses we envision, and solves a ton of potential problems)
* The validity of the read data is checked at elaboration time, via a call to `'Valid` on the read data

Would be interested to have your thoughts

sttaft:
On Tue, Feb 14, 2023 at 9:54 AM Raphaël AMIARD ***@***.***>
wrote:

> Did you mean the "attribute version" above?
>
> Yes
>
> I don't understand this comment. Could you clarify?
>
> What I mean is that we could, I think, define the evaluation of embed in
> such a way that the read is done only once, and every subsequent use is
> only a syntactic "inlining" of the content.
>

But we *never* want it executed on the target machine, so it really needs
to be a special notion of a host-evaluated expression, which is definitely
a bit of an odd beast.


> Furthermore, the version with an aspect does suffer from similar problems,
> if we don't restrict its use to the top level, or at least to constants:
>
> procedure Foo (J : Integer) is
>    Some_Other_Data : Some_Array with External_Initialization => "/some/data.raw";begin
>    Some_Other_Data (J) := 15;
>    Do_Something_With (Some_Other_Data);end Foo;
>
>
This is a bit less of an issue, because aspect specifications are
frequently compile-time or link-time things, with strings that might never
reside on the target, such as the External_Name aspect, and which are
"processed" only once, on the host, during compile time.  Exception
declarations are similarly something that is effectively a compile-time
thing, which are "processed" only once even if the subprogram is executed
hundreds of times.

> ...
>
> That was my sense. Personally I saw many complexities in trying to give
> the attribute version a good definition in Ada such that it was certain to
> be performed on the host machine at compile time, and I am sure Steve Baird
> could come up with an order of magnitude more complexities than me.
>
> We discussed it again, taking your feedback into account, yesterday, and
> discussing other potential topics, such as mutability, constness, where the
> data is located (stack or not), validity of the result, and the version I
> was most satisfied with was something akin to:
>
>    - We use an aspect like External_Initialization in your example above
>    - The use of object declarations with such an aspect is restricted to
>    library level packages (not a big problem for the kind of uses we envision,
>    and solves a ton of potential problems)
>    - The validity of the read data is checked at elaboration time, via a
>    call to 'Valid on the read data
>
> Would be interested to have your thoughts
>
Makes sense to me.  I don't see the strong need to limit it to
library-level objects (though limiting it to library-level types would
definitely help), but it is always better to impose limitations initially
and lift them later, than vice versa.  I would agree that it is more
efficient to only do the validity checks once, and that is more natural for
library-level objects.

Take care,
-Tuck

raph-amiard:
@dkm Could you rewrite the RFC to take the above into account ? TIA :)

sttaft:
In your examples you have:
       `Blah'Valid;`

I think you really want:
      `pragma Assert (Blah'Valid);`

considered/rfc-embed-binary-resources.rst
```diff
@@ -39,27 +39,29 @@ has several issues:
    type Byte is mod 256;
    type Byte_Array is (Integer range <>) of Byte
       with Alignment => 128;
-   Some_File_Data : Byte_Array := Byte_Array'Embed("/some/file/data.raw");
+   Some_File_Data : Byte_Array := External_Initialization ("/some/file/data.raw");
```
jklmnn, line 42: I don't really understand this example. It's not consistent with being an aspect as declared below.

dkm: Sorry, oversight when doing search/replace. Fixed now.

considered/rfc-embed-binary-resources.rst
```diff
@@ -0,0 +1,198 @@
...
+
+  - for constrained type, the number of bytes must match exactly the type size;
+  - for unconstrained type, [FIXME]
+
+.. code-block:: ada
+
+   type Byte is mod 256;
+   type Byte_Array is (Integer range <>) of Byte;
+   Some_File_Data : Byte_Array
+     with External_Initialization (Path =>"/some/file/data.raw",
```
sttaft, line 89: This needs a "=>" *before* the parentheses as well to be syntactically legal:

`    ... with External_Initialization => (Path => ..., Maximum_Size => ...);`

considered/rfc-embed-binary-resources.rst
```diff
@@ -0,0 +1,198 @@
...
+---------------------
+
+On array of scalar type, the ``External_Initialization`` aspect has the same
+observable behavior as if the following transformation is used:
+
+.. code-block:: ada
+
+   type Byte is mod 256;
+   type Byte_Array is (Integer range <>) of Byte;
+   --  Some_File_Data : Byte_Array with External_Initialization ("/some/file/data.raw");
```
sttaft, line 102: This needs a "=>" rather than parentheses:

`    --  Some_File_Data : Byte_Array with External_Initialization => "/some/file/data.raw";`

considered/rfc-embed-binary-resources.rst
```diff
@@ -0,0 +1,198 @@
...
+
+On other type, the aspect has the same observable behavior as in:
+
+.. code-block:: ada
+
+   type Some_Rec_Type is record
+     F1: Integer;
+     F2: Some_Other_Record_Type;
+   end record;
+   --  Some_File_Data : Some_Rec_Type External_Initialization ("/some/file/data.raw");
```
sttaft, line 114: This needs a "with" and a "=>", and the parentheses should be omitted:
`  --   Some_File_Data : Some_Rec_Type with External_Initialization => "/some/file/data.raw";`

jklmnn (approved):

considered/rfc-embed-binary-resources.rst
dkm: Thanks @sttaft , both comments are fixed.

sttaft: Are you sure?  Perhaps I am not seeing the latest version?

considered/rfc-embed-binary-resources.rst
sttaft: Not seeing the update yet online...

jklmnn (dismissed):

considered/rfc-embed-binary-resources.rst
dkm: You're right, thanks for spotting it!

jklmnn (dismissed):

raph-amiard (dismissed):
Minor typo. Apart from that, seems good!

considered/rfc-embed-binary-resources.rst
```diff
@@ -0,0 +1,201 @@
...
+
+A new ``External_Initialization`` aspect is added with the following parameters:
+
+- mandatory ``Path``: the path the compiler uses to access the binary resource;
+- optional ``Maximum_Size``: the maximum number of bytes the compiler reads from
+  the resource;
+- optional ``If_Empty``: an expression used in place of read data in case
+  the resource is empty;
+
+The aspect can only be applied to object declaration at library level.
```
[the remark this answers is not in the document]
dkm: Fixed... And so requires yet-another ACK. I've rebased and squashed at the same time, as next step should be to finally merge it. Thanks!

raph-amiard (approved):

sttaft:
You have used the 'Valid attribute in cases where I believe you should be using the GNAT attribute 'Valid_Scalars.  Ada's 'Valid attribute is not defined for arrays or records (see http://www.ada-auth.org/standards/22aarm/html/AA-13-9-2.html).

t-14:
As a reference, here is a lightweight mixed Ada/Asm example that allows us to achieve similar functionality today.

```
$ cat my.bin
Hello world!
$ cat my_bin.s
.section .rodata
  .global _binary_mybin_start
  .align 4
_binary_mybin_start:
  .incbin "my.bin"
_binary_mybin_end:
  .global _binary_mybin_size
  .align 4
_binary_mybin_size:
  .int  _binary_mybin_end - _binary_mybin_start

$ cat hello.adb
with Text_IO;

procedure Hello is
   Start    : constant Character
     with Import     => True,
          Convention => C,
          Link_Name  => "_binary_mybin_start";
   Length   : constant Integer
     with Import     => True,
          Convention => C,
          Link_Name => "_binary_mybin_size";
   Mybin : String (1 .. Length) with Address => Start'Address;
begin
   Text_IO.Put_Line (Mybin);
end Hello;

$ cat hello.gpr
project Hello is
   for Languages use ("Ada", "Asm");
   for Main use ("hello.adb");
end Hello;

$ gprbuild -q hello
$ hello
Hello world!
```

Presumably, we want to have an intrinsic facility that hides most/all the above book-keeping while providing similar functionality.

In that sense, two things that I'd like to point out:

1. I think restricting it to library-level objects would be unfortunate. As you see, the example above doesn't have this limitation.
2. Given that the proposal involves explicit mention of a file system object in the Ada source code, I am troubled by the associated issues. Requiring an absolute path is certainly a no-no as this would make the source non-relocatable. Supporting a relative path raises the question of "relative to what?" The asm's `.incbin` directive used above relies on `-I` path search mechanism which is not something that Ada currently has. Do we want to make it an "implementation permission" to define how this file object is searched?

Joebeazelman:
I have a hard time seeing a unique use case for this feature. It's useful for localization, but desktop-based OSes , such as MacOS and Windows, already support it through a built-in resource manager. It's is also far more capable, allowing applications to be translated to another region without recompiling the source. 

Perhaps it's intended for command-line programs where the concept of localization is fairly new territory. Swift has a similar proposal, but the implementation might be limited to tools and libraries rather than as a language extension. 

https://github.com/swiftlang/swift-evolution/blob/main/proposals/0278-package-manager-localized-resources.md

The only other use case I can see is a way around Ada's laborious static initialization syntax for large value sets, but this presumes the data is in some easier to compose data format such as CSV or tab delimited, which raises the question if it's a good idea to circumvent Ada's strong typing.

raph-amiard:
Hi @Joebeazelman ,

It's simply useful to embed binary resources directly in your executable, without having to encode it into an Ada data structure. There are already many cases where this is done currently.

* Here is an example from a personal project of mine: https://github.com/raph-amiard/ada-synth-lib/blob/master/src/utils-tan_table.ads
* Another example from the `ada_language_server`: https://github.com/AdaCore/ada_language_server/blob/master/source/ada/generated/lsp-predefined_completion-ada2012.ads

This is particularly useful in embedded contexts where you don't have a filesystem, and where this kind of source embeddings are the only way to embed resources.

But this is also useful in native contexts, because shipping resources with your executable is not always possible/practical.

Joebeazelman:
The method @t-14 describes is precisely how resource files are implemented. They are just data fragments (resources), with a symbol name (the address) in a object file (.o file) linked into the executable (embedded). Alternatively, the object file can be linked as a separate shared library/dynamic library at runtime. 

The differences are resources in Desktop OSes have additional tools for compiling or visually editing them. They also have an API for managing resources, for instance, checking they exist at runtime. The resource or .rc, .res, or .rez files are just typed text representations of the data, similar to JSON which are compiled by the resource compiler into object form and placed in a special 'rsrc' section. During linking, the section is placed in the 'rsrc' section of the application. 

Essentially, Ada and its toolchain already support it, albeit in a cumbersome way. For instance, consider a payload called WIFI containing some connection data in binary form. It can be embedded in the firmware using the linker arguments:

`gcc -Wl,--format=binary -Wl,wifi.bin -Wl,--format=default`

**--format=binary** tells the linker the following file is binary and **--format=default** tells it to switch back to default input format (this is useful if you specify other input files after wifi.bin, such as .o files).

You can access the data from your C firmware code:
```
extern uint8_t data[]     asm("_binary_wifi_bin_start");
extern uint8_t data_end[] asm("_binary_wifi_bin_end");
extern uintptr_t					 asm("_binary_wifi_bin_size);
```
Or from your Ada firmware code:
```
package SSE renames System.Storage_Elements;

Blob : aliased SSE.Storage_Element
     with Import, Link_Name => "_binary_wifi_bin_start";

Length   : constant Integer
     with Import     => True,
          Convention => C,
          Link_Name => "_binary_wifi_bin_size";

WIFI_Data : SSE.Storage_Element (1 .. Length) with Address => Blob'Address;
```

Of course, the application assumes the binary blob is a valid and strongly typed, in this case an array. In a desktop OS, you would access this using a **LoadResource()** and it would return or throw an error if the resource couldn't be found or was invalid.

Fabien-Chouteau:
> Ada and its toolchain already support it, albeit in a cumbersome way

That's the main point of this proposal. Thank @Joebeazelman for highlighting it here.

As for the "desktop OS" use cases, you are going off-topic because you don't take into account the embedded use case. This is, therefore, not at all an alternative to the proposal described here. We kindly ask you to stop arguing in that direction.

If you need more context, I recommend reading this blog post from JeanHeyd Meneide who led the proposal of `#embed` in the C standard: https://thephd.dev/finally-embed-in-c23

Joebeazelman:
Thank you for the clarification. I wasn't sure how this would be implemented. I chimed in because @t-14 made some very good points and they weren't addressed. I brought up resource files because they serve a similar purpose and are incredibly useful. This feature should at least be as powerful and as flexible.

raph-amiard:
> This feature should at least be as powerful and as flexible, i.e. make changes without rebuilding the entire executable.

That's not the point (actually pretty much the opposite), since the point of this feature is 

### Please Put My Data In My Gotdang Executable And Stop Doing Crazy Nonsense, The Feature

(as quoted from the blog post @Fabien-Chouteau linked) :slightly_smiling_face:

PR #105, opened 2023-08-16
[RFC] finally for handled_sequence_of_statements

raph-amiard:
Link to text: https://github.com/AdaCore/ada-spark-rfcs/blob/topic/rfc-finally/considered/rfc-finally.md


Glacia:
>Most languages with exceptions (Java, C#, C++, Python, ...) have a finally block.

C++ doesn't have Finally.

raph-amiard:
Drat, I stumbled upon the one feature C++ doesn't have! :innocent: More seriously, thanks @Glacia, I'll amend the RFC.

filip-gajowniczek-adacore:
For consistency with the other keywords I would recommend the keyword should be either a noun (to match exception block naming) or a verb (to match the begin...end pair of a block).

Since conceptually this seems like a generalization of Finalization, perhaps the keyword "finalize" could make sense.

sttaft:
On Thu, Nov 9, 2023 at 11:11 AM filip-gajowniczek-adacore <
***@***.***> wrote:

> For consistency with the other keywords I would recommend the keyword
> should be either a noun (to match exception block naming) or a verb (to
> match the begin...end pair of a block).
>
> Since conceptually this seems like a generalization of Finalization,
> perhaps the keyword "finalize" could make sense.
>

The big problem with making "finalize" a keyword is that we know it is used
already as the name of a procedure.  Perhaps "finalise" ;-)
Or "finale" which is a nice English noun derived from the French.

-Tuck

raph-amiard:
@filip-gajowniczek-adacore I don't think such consistency exists ("exception" for the exception handler), and "finally" feels pretty natural.

PR #106, opened 2023-10-10
Object Orientation update

QuentinOchem:


considered/rfc-oop-super.rst
```diff
@@ -0,0 +1,171 @@
...
+
+In the case of a type, it refers to its direct parent. E.g.:
+
+.. code-block:: ada
+
+   type T1 is tagged null record;
+
+   procedure P (V : T1);
+
+   type T2 is tagged null record;
```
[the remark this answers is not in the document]
QuentinOchem: Yes

considered/rfc-oop-super.rst
```diff
@@ -0,0 +1,171 @@
...
+   V2 : C2'Super; -- V2 is of type C1
+
+In the case of an object, it refers to the parent of the static type of that
+object. For example:
+
+.. code-block:: ada
+
+  procedure Call (V : T2'Class) is
+  begin
+     V'Super.P; -- static call to T1.V
```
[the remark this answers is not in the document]
QuentinOchem: Agreed.

Glacia:
Someone should make a "Can I Copy Your Homework?" meme for this proposal, considering how much it blatantly copies C++ and Java. 

I understand why this proposal exist (AdaCore customers just want C++ and Java, not Ada), but it really baffles me why would anyone spent so much time on this and not even try to come up with something simpler and better. C++ and Java is definitely not a golden standard in any way, and considering that OOP is slowly going out of fashion (See Rust, Go etc) i would rather see removal of tagged types and OOP stuff from Ada altogether.

Just my two cents.

considered/rfc-oop-primitives.rst
```diff
@@ -0,0 +1,245 @@
...
+
+         procedure P2 (Self : in out T1; V : Integer);
+       end T1;
+   end P;
+
+Class record can still be limited or have discriminants, in which cases the set
+of constaints that they have follow similar rules as for tagged types.
+
+Visibilty rules are the same as for types today. In particular, a class instance
+as access to private components of other instances of the same class.
```
eliericha, line 168: Type `as` -> `has`

considered/rfc-oop-constructors.rst
```diff
@@ -0,0 +1,317 @@
...
+   end P;
+
+As soon as a constructor exist, and object cannot be created without calling one
+of the available constructors, omitting the self parameter. This call is made on
+the object creation, e.g.:
+
+.. code-block:: ada
+
+   V : T1; -- OK, parameterless constructor
+   V2 : T1 (42); -- OK, 1 parameter constructor
```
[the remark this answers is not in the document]
QuentinOchem: So `T1 (42)` isn't supposed to be a value, it's supposed to be a type reference with the parameter needed to call the constructor. 

The equivalence that you describe is interesting. Intuitively, the two lines do not look equivalent to me. In particular, in the second one, you first create an object `T1 (42)`, then you call a copy constructor to create `V`.

QuentinOchem:
@raph-amiard @eliericha following our discussions with regards to constructor syntax, I would like to propose an alternative. How about extending the notation used for initializing an object on the heap? e.g.:

```ada
V1 : Some_Type'(constructor_parameters);
V2 : Some_Type_Access := new Some_Type'(constructor parameters);
```

In V2 case, if "constructor parameters" is an instance of Some_Type, we're already in the right semantic, e.g. this is a case where you want to all a constructor by copy anyway. So in a sense, this becomes "just" an extension of a pre-existing Ada syntax (you already have implicit constructor by copy in the context of heap allocation, you now have constructor by copy in the context of stack allocation and extend to cases where constructors take different parameters).

QuentinOchem:
> @raph-amiard @eliericha following our discussions with regards to constructor syntax, I would like to propose an alternative. How about extending the notation used for initializing an object on the heap? e.g.:
> 
> ```ada
> V1 : Some_Type'(constructor_parameters);
> V2 : Some_Type_Access := new Some_Type'(constructor parameters);
> ```
> 
> In V2 case, if "constructor parameters" is an instance of Some_Type, we're already in the right semantic, e.g. this is a case where you want to all a constructor by copy anyway. So in a sense, this becomes "just" an extension of a pre-existing Ada syntax (you already have implicit constructor by copy in the context of heap allocation, you now have constructor by copy in the context of stack allocation and extend to cases where constructors take different parameters).

Following live discussions, it seems that this syntax is promissing, and has the nice property of avoiding confusion with conversions. 

However, the syntax 

```ada
V1 : Some_Type'(constructor_parameters);
```

Introduces a number of problems from a language design perspective (it is bizarre to mix both a variable type declaration and a call to a constructor. This is not common).

The proposal for the above is to have the user write instead:

```ada
V1 : Some_Type := Some_Type'(constructor_parameters);
```

The exact interaction with aggregates isn't entirely clear yet, but that's a potential path to the right direction. Updating the proposal to reflect that.

considered/rfc-oop-obsolete.rst
```diff
@@ -0,0 +1,50 @@
...
+other than the first, or dispatching on results. If you declare primitives with
+references to the type other than the first parameter, they will not be used
+for
+
+Under the current model, coextensions are replaced by constructors
+(it's possible to mandate an object to be used in the construction of the
+class) and destructors (that same object can always be destroyed in the
+destructor). There is no way to create a coextension on a class record.
+
+Controlled types are replaced by constructor / destructors and are not allowed.
```
raph-amiard, line 31: This is not expanded upon here, but you also talk about it in the rfc-oop-tagged document, mentioning that controlled components are not allowed.

This seems to me like being way too restrictive and unnecessary for the moment.

* First problem is that many facilities that we provide to users use controlled types. If you cannot store a vector in a class, we will have a big problem.
* Second, we don't have a replacement for the `Controlled` mechanism yet, and when we do, it seems very likely that it will *replace* and include the controlled model rather than be an additional different mechanism.

considered/rfc-oop-constructors.rst
```diff
@@ -0,0 +1,332 @@
...
+         procedure T1 (Self : in out T1);
+         procedure T1 (Self : in out T1; Some_Value : Integer);
+      end T1;
+
+      type T2 is record
+         procedure T2 (Self : in out T2; Some_Value : Integer);
+      end T2;
+   end P;
+
+As soon as a constructor exist, and object cannot be created without calling one
```
raph-amiard, line 37: typo: *an

considered/rfc-oop-final.rst
```diff
@@ -0,0 +1,129 @@
...
+Final classes
+-------------
+
+tagged record and class record also implement the concept of final classes,
+which is a class not deriveable. There are two advantages of final classes:
+
+- In terms of design, this makes it clear that this class is not intended to be
+  derived. It's often the case where derivation is used just to have a class in
+  a given framework but isn't prepared to be itself modified.
+- A very significant one: a final class is effectively a definite type.
```
raph-amiard, line 54: You should link to the corresponding RFC here

considered/rfc-oop-final.rst
```diff
@@ -0,0 +1,129 @@
...
+
+Other top level primitives may be needed here.
+
+However, there are several elements that argue against this design:
+
+- the language that implement that (Java) initially introduced that as a way
+  to workaround lack of genericity and `void *` notation. Ada provides
+  genericity, and in the extreme cases where `void *` is required,
+  `System.Address` is a reasonable replacement.
+- As opposed to Java, many types in Ada are not objects. This concept would then
```
raph-amiard, line 110: As discussed live, even Java has exceptions to this (`int` vs. `Integer for example)

considered/rfc-oop-first-controlling.rst
```diff
@@ -0,0 +1,67 @@
...
+=======
+
+
+Motivation
+==========
+
+Guide-level explanation
+=======================
+
+A new pragma / aspect is introduced for tagged types, "First_Controlling_Parameter"
```
raph-amiard, line 16: This does seem indeed to be a very good superfeature for what we needed in LAL!

sttaft: It would be helpful to indicate how this differs from using class-wide types for parameters or results that are not supposed to be dispatching.   Also, this presumably is not legal for an abstract type.

One alternative approach would be to have a "special child" of a given root type that is not generally further extended.  The extensions of the root type can have parameters and results of this "special child" without making those parameters/results dispatching.

It would certainly help to see some use cases where you want to continue to refer to the root type all the way down the type hierarchy.  Note that in Java in general, and in C++ when using "ref" parameters, the parameters are effectively "class-wide" in that the actual run-time type might not match the declared type of the parameter or result.   That makes the use of a class-wide parameter/result a better match for Java and C++ in such contexts.  Continuing to refer only to the root type specifically is not something that happens at all in Java, and happens in C++ only when passing a parameter by copy, so it is interesting that there are important use cases for this in Ada.

raph-amiard: > It would be helpful to indicate how this differs from using class-wide types for parameters or results that are not supposed to be dispatching. Also, this presumably is not legal for an abstract type.


@sttaft in case it's not obvious, because it allows returning definite types, and in general simplifies the very general use case.

But I agree that it should be indicated @QuentinOchem

sttaft: I guess I could still use help envisioning when a function for a derived type would want to return a value of its ancestor type.  Can you give some examples (other than a simple conversion function which would be an identity in the root type)?

QuentinOchem: > > It would be helpful to indicate how this differs from using class-wide types for parameters or results that are not supposed to be dispatching. Also, this presumably is not legal for an abstract type.
> 
> @sttaft in case it's not obvious, because it allows returning definite types, and in general simplifies the very general use case.
> 
> But I agree that it should be indicated @QuentinOchem

Done :-)

considered/rfc-oop-constructors.rst
```diff
@@ -272,7 +312,7 @@ declaring such parameters:
          procedure Some_T (Self : in out Some_T; X, Y : Integer);
       end Some_T;
    package G
-      X : Some_T (5, 6); -- OK, we expect a 2 parameters con
+      X : Some_T := Some_T'(5, 6); -- OK, we expect a 2 parameters con
```
eliericha, line 315: I have a feeling that we could do without passing an expected constructor as part of the generic formal. That feeling stems from my experience with Java where a generic class can't instantiate an object of its generic type parameter. it can handle such objects, pass them around, but I don't think it can instantiate such an object itself.

So I wouldn't be surprised if Ada had the same constraint.

Moreover, the explicit formal constructor could be replaced with a formal factory function that constructs an object using the desired construction parameters , without the need for dedicated syntax.

For example (with approximate syntax because I don't know the syntax for generics well):

```ada
generic
   type Some_T is new T2;
   function Create_Some_T (X, Y : Integer) returns Some_T;
package G
   X : Some_T := Create_Some_T (5, 6);
```

considered/rfc-oop-first-controlling.rst
sttaft: On further thought, I now suspect what you want are *nonprimitve* functions that return the type of interest, say "T".  I was earlier presuming you wanted primitive functions (meaning the first parameter was of our type T) that returned T, but on inheritance the return type would stay as T while the first parameter shifted to T2.  But if the function returns T but is not to be primitive, one easy approach is to define a nested or child package called, say, Factory, and put into the Factory subpackage, all such functions or other operations that are about building objects of type T, but which are not to be inherited.

CodePeer does this pretty systematically, and has child packages called Factory where "constructor" functions live.

considered/rfc-oop-super.rst
```diff
@@ -0,0 +1,158 @@
...
+
+   type T2 is new T1 with null record;
+
+   procedure P (V : T2);
+
+   type C1 is class record
+      procedure P (V : C1);
+   end C1;
+
+   type C2 is class record
```
eliericha, line 37: I think this is missing an extension of C1.

considered/rfc-oop-constructors.rst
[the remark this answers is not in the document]
eliericha: I still don't understand how `T1 (42)` was a type reference when `42` was the parameter to the constructor. But it's OBE if we go for the new syntax ``T1'(42)`` anyway.

considered/rfc-oop-constructors.rst
```diff
@@ -138,7 +150,7 @@ anymore (with the exception of the copy constructor):
       end T1;
    end P;
 
-   V : T1 (10); -- illegal
+   V : T1'(10); -- illegal
```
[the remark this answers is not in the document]
eliericha: OBE as this is part of a more general discussion on the interaction between discriminants and constructors.

considered/rfc-oop-constructors.rst
eliericha: Although the content changed, this comment is still relevant.

Now pushing the thread a bit more, my suggestion of not declaring the constructors can be a problem when extending the type within the generic:

```ada
generic
   type Some_T is new T2;
package G
   type Child_T is new Some_T with record
      procedure Child_T (Self : in out Child_T)
      with Super (1, 2)
      is ...
```

At the `Super` call how does the compiler know that this constructor is available? Either it must be declared as in the initial proposal, or the compiler can determine that at the point of instantiation of the generic when it has visibility over the actual type. I like the latter for simplicity of the declaration but I realize that it breaks modular compilation because we lack information when compiling the generic alone - I don't know if that's a real blocker or if there are already precedents.

Note that Java avoids that problem by disallowing the extension of generic type parameters:

```java
class Gen<T> {
   class Inner extends T { // Cannot refer to the type parameter T as a supertype
   }
}
```

considered/rfc-oop-constructors.rst
```diff
@@ -198,72 +297,130 @@ for example:
...
+exiting the constructor. In particular, the initializion list is not expected
+to create a predicate-valid type - perdicate will only be checked after the
+constructor has been processed.
 
-These considerations are applicatble to both class records and simple records.
+Constructors And Aggregates
+---------------------------
 
 Ada 2022 already allows homogeneous data structure aggregates to be expressed
 through angular brackets. This proposal extends that notation to hetoregeneous
```
eliericha, line 426: The sentence introduces an example about heterogeneous data structures but the example is a homogeneous data structure.

considered/rfc-oop-constructors.rst
```diff
@@ -285,54 +442,72 @@ In the presence of constructors, aggregates values are evaluated and assigned
...
+
+      procedure R2 (Self : in out R2);
+   end record;
+
+   V1 : R1 := [1, 2]; -- prints Default Default
+   V2 : R2 := [1, 2]; -- also prints Default Default
+
+This means that units compiled with the new version of Ada will have a specific
+backward incompatible change. This could be identified statically by migration
+tools.
```
eliericha, line 475: I suggest elaborating what the old rule that is being broken is. Is it that initialization with an aggregate bypasses default initialization?

considered/rfc-oop-constructors.rst
```diff
@@ -285,54 +442,72 @@ In the presence of constructors, aggregates values are evaluated and assigned
...
+
+In terms of syntax, in the presence of an implicit or explicit default
+constructors, aggregates can be written as usual. The default constructor will
```
eliericha, line 478: I'm realizing now that the term "default constructor" is not well defined in the current proposal.

We use the term "default constructor" to mean a parameterless explicitly declared constructor, while we also use the term "default copy constructor" to mean an implicit undeclared constructor.

For some context, Java uses the term "default constructor" to denote an implicit undeclared constructor that is provided by the compiler when no explicit constructors are declared. Of course we don't have to use the same terminology but we should define one more clearly.

considered/rfc-oop-constructors.rst
```diff
@@ -385,28 +558,91 @@ constructors are provided. For example:
...
+      V : T := Create (55);
+   end G;
+
+   package P is
+
+      type T2 is tagged record
+         procedure T2 (Self : in out T1; V : Integer);
+      end record;
+
+      package G2 is new G (T2, T2'Make); -- Legal
```
eliericha, line 635: I assume this would also work in the presence of multiple constructors, and the one that matches the formal signature is automatically resolved? Is it possible to have ambiguity? If yes, how would the ambiguity be resolved?

I think we should piggy back on existing rules for resolving actual subprograms to formal generic parameters which probably already answer these questions. But may be worth mentioning it here or in the other section where the matter of referencing constructors as functions was presented.

considered/rfc-oop-constructors.rst
```diff
@@ -0,0 +1,719 @@
...
+Rationale and alternatives
+==========================
+
+Rationale for Initialization Lists
+----------------------------------
+
+Languages like Java or Python do not require initialization lists. However, by
+default, class fields are references and initialized by null. In system-level
+languages like C++ or Ada, we want to be able to have fields as direct members
+of their enclosing records (as opposed to references). However, these tagged
```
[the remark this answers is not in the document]
QuentinOchem: yes

considered/rfc-oop-fields.rst
```diff
@@ -0,0 +1,92 @@
...
+==========
+
+Guide-level explanation
+=======================
+
+Components declarations
+-----------------------
+
+A record, tagged record and class record type can now have 3 declarations
+region where fields can be declared:
```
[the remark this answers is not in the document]
QuentinOchem: Yes

considered/rfc-oop-final.rst
```diff
@@ -0,0 +1,129 @@
...
+changed after the constructor call, not even during aggregate which is
+considered as a shortcut for assignment. For example:
+
+.. code-block:: ada
+
+   package P is
+      type T1 is class record
+         procedure T1 (Self : in out T1; Val : Integer);
+
+	      Y : final Integer := 0;
```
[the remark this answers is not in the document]
QuentinOchem: Nope, tabs instead of spaces, fixing.

considered/rfc-oop-final.rst
```diff
@@ -0,0 +1,129 @@
...
+      end T1;
+   end P;
+
+   package body P is
+      type body T1 is class record
+         procedure T1 (Self : in out T1; Val : Integer) is
+	      begin
```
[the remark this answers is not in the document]
QuentinOchem: Same issue

considered/rfc-oop-primitives.rst
```diff
@@ -0,0 +1,245 @@
...
+            Self.F := V + 1;
+         end P2;
+       end C;
+
+   end P;
+
+Primitives declared within a type can only be called via prefix notation. In
+the specific case of a class record, subprogram declared outside of the lexical
+scope are not primitive (they can't be called via prefix notation, they can't
+be overriden).
```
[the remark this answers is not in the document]
QuentinOchem: At this stage of the discussion, class record are a very different kind of declaration when it comes to prefix notation. To some respect, we did need to generalize prefix notation to types because we could not declare operations within their scope. So presumably, the only prefixable operations of a class are those declared in its scope. That's still up for discussion though.

considered/rfc-oop.rst
```diff
@@ -0,0 +1,128 @@
+- Feature Name: Standard OOP model
+- Start Date: May 5th 2020
+- RFC PR:
+- RFC Issue:
+
+Summary
+=======
+
+The objective of this proposal is to draft an OOP model for a hypothetical new version of Ada, closer to the models languages
+such as C++ and Java developers are accustomed to, fixing a number of oddities and vulnerabilities of the current Ada language.
```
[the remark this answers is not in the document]
QuentinOchem: There's a few examples: the fact that in current Ada, there's no way to enforce that a subprogram is called at creation (constructor) or destruction (destructor) of an object is a potential vulnerability. You can use Controlled types but (1) you can shortcut Initialize in various contexts and (2) you can't force call the Initialize / Finalize parents in a child class. Another issue relates to the fact that dispatching isn't the default calling mode in Ada OOP, making it easy to mistakenly call a primitive that is not the one that has been overridden in the actual type of the object, thus calling an operation that may leave the object in an inconsistent state.

considered/rfc-oop-constructors.rst
```diff
@@ -91,97 +129,158 @@ data:
...
-Initialization of these objects can be done with the `Initialize` aspect. For
-example:
+The following sections will describe all three cases.
+
+Initialization of Components
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Initialization of these components can be done with the `Initialize` aspect.
+This is particularly useful when component types do not have default
+constructors, but is generally available. For example:
```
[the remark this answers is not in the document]
QuentinOchem: Will adapt the text indeed.

Note that you can depend on parameters on default component initialization, but these parameters needs to be global ones while often you'll want these parameters to depend on the constructor parameters.

considered/rfc-oop-constructors.rst
```diff
@@ -198,72 +297,130 @@ for example:
...
+         with Initialize (Super => Default_Root)
+      is
+      begin
+         null;
+      end Child;
+   end Child;
+
+In addition, initializing the super is the only place where a value of an
+abstract type can be created, as it will be completed by a concrete type. For
+example:
```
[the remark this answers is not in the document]
QuentinOchem: Indeed, rewriting to avoid the "created" wording.

considered/rfc-oop-constructors.rst
QuentinOchem: Yup, elaborating.

considered/rfc-oop-constructors.rst
```diff
@@ -440,3 +687,33 @@ Unresolved questions
...
+   package P is
+      type T1 (<>) is tagged record -- T1 is indefinite
+	      X : String;
+
+         procedure T1 (Size : Ineger);
+      end record;
+
+      type body T1 (<>) is tagged record
+         procedure T1 (Size : Integer)
+            with Initialize (X => String'(1 .. Size));
```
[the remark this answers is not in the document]
QuentinOchem: Indeed. Fixing.

considered/rfc-oop-constructors.rst
```diff
@@ -0,0 +1,731 @@
...
+      end T1;
+   end P;
+
+Reference-level explanation
+===========================
+
+Rationale and alternatives
+==========================
+
+Rationale for Initialization Lists
```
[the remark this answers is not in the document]
raph-amiard: On second thought, I really like the perspective you open up in "Record with Indefinite Fields". This alone kind of makes the initializer lists proposal worthwile IMO

considered/rfc-oop-attributes.rst
```diff
@@ -0,0 +1,80 @@
...
+      Item : in T);
+
+This proposal allows to define directly the procedure as being an attribute of
+the type:
+
+.. code-block:: ada
+
+   type T is null record;
+
+   procedure T'My_Write(
```
mgrojo, line 35: Shouldn't this be `T'Write`?

QuentinOchem: Absolutely - thanks for the review :-)

PR #115, opened 2024-05-30
OOP - Constructors

QuentinOchem:
This RFC contains specific elements about the new constructor syntax, coming from the overall OOP discussion.


ebotcazou:
The canonical moniker is "untagged" in Ada, so "tagged and untagged records".

QuentinOchem:
> The canonical moniker is "untagged" in Ada, so "tagged and untagged records".

Thanks - fixed.

PR #118, opened 2024-05-30
[RFC] OOP - Dispatching 

QuentinOchem:
This RFC contains specific elements about the new dispatching syntax and semantics, coming from the overall OOP discussion.

### DIFF
--- a/considered/rfc-oop-dispatching.rst
+++ b/considered/rfc-oop-dispatching.rst
@@ -45,6 +45,59 @@ explicitely marked as static (for example, through 'Super). E.g:
 
 Default_Dispatching_Calls is On by default on new version of the language.
 
+Multi-Parameter Dispatching
+---------------------------
+
+Under this new pragma Default_Dispatching_Calls, primitives with more than
+one controlling parameter behave in the following way:
+
+- If they're called on a dispatching call, then everything works as if all
+  controlling parameters were converted to class wide view. And indeed, in that
+  case, we would have dynamic tag check as you do today in Ada.
+
+- If it's a non-dispatching call, today that's through 'Super, then you will
+  statically select the primitive, and will need to be able to statically check
+  the static type of all parameters.
+
+For example:
+
+.. code-block:: ada
+
+   type Root is tagged null record;
+
+   procedure Prim (A, B : Root);
+
+   type Child is new Root with null record;
+
+   overriding procedure Prim (A, B : Child);
+
+   R1, R2 : Root;
+   C1, C2 : Child;
+
+   C1'Super.Prim (R2); -- static, legal
+   C1'Super.Prim (C2'Super); -- static, legal
+   C1'Super.Prim (C2); -- illegal, C2 is of the wrong type
+
+Note that this is a problem when integrating with current Ada, pedantic Flare
+does not support multi-parameter dispatching.
+
+Dispactching on Returned Types
+------------------------------
+
+A tag indeterminate disatching call is illegal (as it is the case today). For
+example:
+
+.. code-block:: ada
+
+     pragma Default_Dispatching_Calls;
+     type T is tagged ... ;
+     function Make return T; -- primitive
+     Obj1 : T'Class := ...
+     Obj2 : T'Class := Make; -- illegal
+   begin
+     Obj1 := Make; -- legal; use Obj1'Tag to dispatch
+
+
 Reference-level explanation
 ===========================
 
--- a/considered/rfc-oop-dispatching.rst
+++ b/considered/rfc-oop-dispatching.rst
@@ -1,0 +1,97 @@
+- Feature Name: Standard OOP model
+- Start Date: May 5th 2020
+- RFC PR:
+- RFC Issue:
+
+Summary
+=======
+
+Motivation
+==========
+
+Guide-level explanation
+=======================
+
+Dispatching and Class Wide Views
+--------------------------------
+
+A new library / partition level pragma is introduced, Default_Dispatching_Calls.
+When active, calls to primitives are dispatching (when the tag can be
+determined statically, the compiler may optimize the call to be static) unless
+explicitelly marked as static (for example, through 'Super). E.g:
+
+.. code-block:: ada
+
+   pragma Default_Dispatching_Calls (On);
+
+   type Root is tagged null record;
+
+   procedure P (Self : in out Root);
+
+   type Child is new Root with null record;
+
+   procedure P (Self : in out Root);
+
+   type A_Root is access all Root;
+
+   procedure P (V : in out Root) is
+   begin
+      V.P2; -- Dispatching
+   end P;
+
+   V : A_Root := new Child;
+
+   V.P; -- dispatching
+
+Default_Dispatching_Calls is On by default on new version of the language.
+
+Reference-level explanation
+===========================
+
+
+Rationale and alternatives
+==========================
+
+It is a potential vulnerability not to call an overriden primitive. This may
+lead to an object to be in an state that has not been anticipated, in particular
+when the role of the overriden primitive is to keep the state of the derived
+object consistant. It's also commonly the case in most OOP languages that
+dispatching is the default expected behavior and non dispatching the exception.
+
+This also fix a common confusion in Ada, where the dispatching parameter of A
+primitive is itself non-dispatching and requires so-called redispatching. The
+following code becomes then much more natural to write:
+
+.. code-block:: ada
+
+   package P is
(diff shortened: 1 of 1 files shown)


considered/rfc-oop-dispatching.rst
```diff
@@ -0,0 +1,97 @@
...
+Guide-level explanation
+=======================
+
+Dispatching and Class Wide Views
+--------------------------------
+
+A new library / partition level pragma is introduced, Default_Dispatching_Calls.
+When active, calls to primitives are dispatching (when the tag can be
+determined statically, the compiler may optimize the call to be static) unless
+explicitelly marked as static (for example, through 'Super). E.g:
```
tonunaks, line 21: explicitly

considered/rfc-oop-dispatching.rst
```diff
@@ -0,0 +1,97 @@
...
+Rationale and alternatives
+==========================
+
+It is a potential vulnerability not to call an overriden primitive. This may
+lead to an object to be in an state that has not been anticipated, in particular
+when the role of the overriden primitive is to keep the state of the derived
+object consistant. It's also commonly the case in most OOP languages that
+dispatching is the default expected behavior and non dispatching the exception.
+
+This also fix a common confusion in Ada, where the dispatching parameter of A
```
tonunaks, line 61: This also **fixes** a common confusion in Ada, where the dispatching parameter of **a** ...

considered/rfc-oop-dispatching.rst
```diff
@@ -0,0 +1,97 @@
...
+primitive is itself non-dispatching and requires so-called redispatching. The
+following code becomes then much more natural to write:
```
tonunaks, line 63: This formulation makes me thing that the following code is "as-is" example followed by the new improved version. I suggest adding an example of the current confusing version and then the code that simplifies it

considered/rfc-oop-dispatching.rst
QuentinOchem: Fixed.

considered/rfc-oop-dispatching.rst
QuentinOchem: Fixed

considered/rfc-oop-dispatching.rst
QuentinOchem: Indeed - and the following syntax is assuming scoped primitives which is not confirmed just yet. Modified the text and the example to make that more clear.

clairedross:
Very interesting. Do we really want to set the mode by default though ? This does not seem very user friendly to me.

clairedross:
As discussed live with Quentin, I am worried dispatching everywhere might remove the possibility to support OOP in SPARK. Basically, SPARK support of OOP is already quite rudimentary. It does not support many design patterns, because basically, when it proves something, it tries to prove it for all possible derived types with all possible overriding primitives. So what you can do is (1) prove subprograms using dispatching for all possible derived types using the provided classwide contracts (2) prove specific subprograms where the tag is known for the object. These two usages are made possible by the specific handling of calls to primitive operations in classwide contracts in Ada. They are considered as dispatching calls on dispatching calls, which allows to handle (1), but when proving the contract itself, they call the function of the specific type (which changes when they are inherited) so SPARK can prove (2). Here is an example:

```
package P with SPARK_Mode is
   pragma Elaborate_Body;

   type Root is abstract tagged null record;
   function Is_Init (R : Root) return Boolean is abstract;
   function Is_Processed (R : Root) return Boolean is abstract;
   procedure Init (R : out Root) is abstract with
     Post'Class => Is_Init (R);
   procedure Process (R : in out Root) is abstract with
     Pre'Class  => Is_Init (R),
     Post'Class => Is_Processed (R);
   procedure Go (R : out Root) with
     Extensions_Visible,
     Post'Class => Is_Processed (Root'Class (R));

   type Child is new Root with record
      F : Integer;
   end record;
   function Is_Init (R : Child) return Boolean is (R.F >= 1);
   function Is_Processed (R : Child) return Boolean is  (R.F >= 0);
   procedure Init (R : out Child);
   procedure Process (R : in out Child);
end P;

package body P with SPARK_Mode is

   procedure Go (R : out Root) is
   begin
      Init (Root'Class (R));
      Process (Root'Class (R));
   end Go;

   procedure Init (R : out Child) is
   begin
      R.F := 1;
   end Init;

   procedure Process (R : in out Child) is
   begin
      R.F := R.F - 1;
   end Process;

end P;
```

The procedure Go can be proved because calls in the contracts of Init and Process are considered to be dispatching and the procedures Init and Process can be proved because they see the specific versions. Note that I wrote Go as a primitive here, just so we can see the explicit redispatch, but it would make more sense to use a parameter of type Root'Class as it is probably not meant to be overridden.

As I think I demonstrated, the special handling in classwide contracts is crucial for provability of OOP in SPARK. Arguably, the support of OOP in SPARK is already not that great, but I think it might still be worthwhile to try and salvage what we have. Requiring the use of "old" OOP for SPARK would still be possible of course, but it might make the gap between Ada and SPARK wider while we have been trying for years to reduce it.

sttaft:
The approach that existing Ada takes where dispatching only occurs when parameters are of a class-wide type, gives the programmer control of where and when the level of indirection and dynamism of dispatching is used.  A dispatching call is a more challenging thing to test and prove things about than a statically-bound call, so it should be a conscious choice. 

Historically, Ada's approach was developed in response to concerns expressed by organizations (like Microsoft) who were supporting large, heavily object-oriented libraries (such as Microsoft Foundation Classes).  They found that any use of dispatching calls from one primitive of a type to another created complex coupling.  This coupling meant that if some but not all primitives were inherited by a derived type (subclass), then the details of the inter-primitive calls made by the inherited primitives, that might normally be considered private to the implementation of the type, become visible.  That is, the user would be relying on an undocumented internal re-dispatch, from say an inherited "Print_Line" operation to an overridden "Print" primitive.  Once users figure out these inter-primitive calls, they never want to see them change on new versions of the library, an expectation which could make maintenance quite difficult.

We recognized that such inter-primitive re-dispatching calls need to be documented, and hence should not be the default.  For a case like Print_Line, in Ada the simplest solution is to have Print_Line take a class-wide parameter in the first place, which it then uses to dispatch to Print and New_Line, if that is the desired semantics.  Print_Line then becomes a "class-wide operation" rather than a primitive, eliminating the potential confusion and the dangers of partial inheritance.

manthonyaiello:
Let’s build on Claire’s comment above to make sure the proposed change to always dispatch won’t break not only SPARK, but potentially any Ada that uses postconditions on classes.

We’ll change Claire’s code as follows, and assume that this RFC has been implemented (i.e., we always dispatch, i.e., the ‘self’ parameter is always converted to a classwide).

```ada
package P with SPARK_Mode is
   pragma Elaborate_Body;

   type Root is abstract tagged null record;
   function Is_Init (Self : Root) return Boolean is abstract;
   procedure Init (Self : out Root) is abstract with
     Post => Self.Is_Init;

   type Child is new Root with record
      F : Integer;
   end record;
   function Is_Init (Self : Child) return Boolean is (Self.F >= 1);
   procedure Init (Self : out Child);

   type Grandchild is new Child with record
      G : Integer;
   end record;
   function Is_Init (Self : Grandchild) return Boolean is (Self'Super.Is_Init and Self.G >= 2);
   procedure Init (Self : out Grandchild);
end P;

package body P with SPARK_Mode is

   procedure Init (Self : out Child) is
   begin
      Self.F := 1;
   end Init;

   procedure Init (Self : out Grandchild) is
   begin
      Self'Super.Init;  -- This line results in a failed postcondition
      Self.G := 2;
   end Init;
end P;
```

The problem arises at the call to `Self’Super.Init;` in `Grandchild`. At the conclusion of `Child.Init`, the postcondition is evaluated and `Is_Init` is called. However, since we always dispatch and since `Grandchild` overrides `Is_Init`, we will call `Grandchild.Is_Init` at this point - which will be false, so the postcondition will fail.

It’s potentially problematic in general to call a superclass method that may call an overridden method - but there are ways to protect against it when needed (final, static, etc.). Here the way out of the box is not clear:
* we don’t want to mark `Init` as final
* we don’t want the call to `Is_Init` to be never-dispatching, because used outside of the postcondition, the dispatching behavior is likely to be expected.

Perhaps the solution is to make the postcondition on `Root.Init` specify non-dispatching?

```ada
   procedure Init (Self : out Root) is abstract with
     Post => Self.Is_Init'Static;  -- (or some other syntax)
```

tonunaks (approved):

PR #120, opened 2024-09-11
[RFC] OOP Components

QuentinOchem:
This RFC contains specific elements about the new component declaration capabilities, coming from the overall OOP discussion.


considered/rfc-oop-fields.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Components declarations
+-----------------------
+
+A record and tagged record type can now have 3 declarative regions where fields
+can be declared:
+
+- the package public view
+- the package private view
+- the record private view
```
tonunaks, line 23: Does the "record private view" refer to the split shown below where type is marked as "with private" and then elements are added in the package private section or do you expect that the record declaration itself contains section labelled "private"?

QuentinOchem: That's a leftover from a part of the proposal that was carved out (the so-called "protected" region). In this proposal, there is indeed only two sections. Let me fix that.

PR #124, opened 2024-09-11
[RFC] Default short-circuit operators

QuentinOchem:
This RFC proposes to have predefined boolean operators "and" and "or" to be short-circuit by default.

### DIFF
--- a/considered/rfc-shortcircuit.md
+++ b/considered/rfc-shortcircuit.md
@@ -27,10 +27,13 @@ evaluation is guarded by the left handside, for example:
 Guide-level explanation
 =======================
 
-We introduce a new pragma Shorcircuit_Operators which is a library-level pragma
-governing the semantic of boolean operator in the unit where the pragma is used.
-For code compiled until Ada 2022, the value of this pragma is False, True
-afterwards.
+Overall Concept
+---------------
+
+We introduce a new pragma `Shorcircuit_Operators` which is a library-level
+pragma governing the semantic of boolean operator in the unit where the pragma
+is used. For code compiled until Ada 2022, the value of this pragma is Off,
+On afterwards.
 
 When the pragma is in effect and has a true value, non-overriden `and` and `or`
 operators are now shortcircuit. More specifically:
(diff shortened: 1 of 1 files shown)


sttaft:
This pragma could create a significant number of subtle changes in run-time behavior if ever applied to existing code, and will also mean that long-time readers of Ada code may misinterpret the runtime semantics of code using "and" and "or" operators.  Furthermore, the standard transformation of `X op Y` to `"op"(X, Y)` will no longer work for "and" (and "or") operators, because function parameter evaluation does not follow short-circuit semantics.  Parameters X and Y will both be evaluated in `"and"(X, Y)` even though `X and Y` would only evaluate Y if X evaluated to True.

A safer and simpler alternative would be defining a new restriction ID for pragma Restrictions that would disallow use of "and" and "or" operators on boolean types.  This way a project could avoid any potential errors associated with use of non-short-circuit `and`/`or`, but perhaps would still allow the use of `"and"(X, Y)` in cases where non-short-circuit evaluation is considered important for, say, [constant-time computation](https://www.chosenplaintext.ca/articles/beginners-guide-constant-time-cryptography.html) reasons in cryptography, or performance when the operands are simple boolean flags.

manthonyaiello:
Making all boolean operators short-circuit by default and not offering a non-short-circuit alternative will have a potential impact on the expressibility of certain contracts. The Ada RM forbids the use of `'Old` in conditional contexts because the compiler evaluates the '`Old' when a subprogram is called (6.1.1 27.g and h/5).

Here's the rationale:

> Reason: {[AI05-0273-1](http://www.ada-auth.org/cgi-bin/cvsweb.cgi/AI05s/AI05-0273-1.TXT)} {[AI12-0280-2](http://www.ada-auth.org/cgi-bin/cvsweb.cgi/AI12s/AI12-0280-2.TXT)} {[AI12-0005-1](http://www.ada-auth.org/cgi-bin/cvsweb.cgi/AI12s/AI12-0005-1.TXT)} Since the [prefix](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-4-1.html#S0093) of an Old [attribute_reference](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-4-1-4.html#S0100) is evaluated when the subprogram is called (if it is evaluated at all), we cannot allow it to include values that do not exist at that time (like 'Result and loop parameters of [quantified_expression](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-4-5-8.html#S0153)s). We also do not allow the [prefix](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-4-1.html#S0093) itself to include 'Old references, as those would be redundant (because the evaluation of the [prefix](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-4-1.html#S0093), if it occurs, already happens on entry to the subprogram ), and allowing them would require some sort of order to the implicit constant declarations (because in A(I'Old)'Old, we surely would want the value of I'Old evaluated before the A(I'Old) is evaluated).
>
> {[AI05-0273-1](http://www.ada-auth.org/cgi-bin/cvsweb.cgi/AI05s/AI05-0273-1.TXT)} {[AI12-0280-2](http://www.ada-auth.org/cgi-bin/cvsweb.cgi/AI12s/AI12-0280-2.TXT)} An additional rule applies when it cannot be determined on entry to the subprogram whether the Old [attribute_reference](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-4-1-4.html#S0100) will or will not be evaluated when the overall postcondition expression is evaluated. In such cases, we require that the [prefix](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-4-1.html#S0093) of the Old [attribute_reference](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-4-1-4.html#S0100) to statically name some object. This is necessary because the Old [prefix](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-4-1.html#S0093)es have to be evaluated when the subprogram is called if there is any possibility that they might be needed; the compiler cannot in general know whether they will be needed in the postcondition expression.

tonunaks:
I am approving this text with the understanding that the details for gnat release 26 still need to be refined. We agreed to merge it as is, and the FE team (in person of @bobduff) will propose refinements in a separate pull request.

Issue #126, opened 2024-11-04
[feature] More precise “modes” for data modified by a subprogram


clairedross:
### Summary

A new contract could be used to precisely express which parts of a parameter/global of mode IN OUT are modified/preserved by a subprogram. It would be interesting to also allow expressing in which cases these (parts of) objects are preserved. 

```ada
   type R is record
      F1, F2 : Boolean;
   end record;

   procedure Write_F1 (X : in out R) with
     Modifies => (others => X.F1);

   procedure Write_F1 (X : in out R) is
   begin
      X.F1 := False;
   end Write_F1;

   G : Integer;

   procedure Write_G_If_B (B : Boolean) with
     Global => (In_Out => G),
     Modifies => (B => G);

   procedure Write_G_If_B (B : Boolean) is
   begin
      if B then
         G := 0;
      end if;
   end Write_G_If_B;
```
More generally, this contract could be structured like a Contract_Cases with various cases made of a precondition and a sequence of paths into parameters/globals of mode IN OUT. Paths here could be defined like for deep delta aggregates.

### Motivation

Modes for parameters and global variables modified by a subprogram are sometimes too coarse, it would be useful to easily express which parts of the data are preserved or modified in a contract.

### Caveats and alternatives

This contract is theoretically equivalent to a contract cases where the consequences could be written as `X = (X'Old with delta Path => X.Path'Old ...)`. It can be handled that way for proof, but it is more complex for execution as the equality here is not the Ada equality but the logical one, so this contract would probably remain non-executable (like Global and Depends).

An alternative would be to rather list preserved parts using a `Preserves` annotation.

clairedross:
> "Modifies" seems more maintainable than "Preserves" in that the number of "Modifies" correlates with the number of updates performed, while "Preserved" goes just the opposite. With "Preserved", the moment you update a single field, you need to specify the full list of other fields in a "Preserves", which seems burdensome and probably too noisy.

I agree with that. I also like "Modifies" better for tagged types as hidden components are preserved by default.

> I do wonder what you expect to happen with a private type. Ada 2022 has the "Stable_Properties" aspect which is more like Preserved, but is generally at the type level, but can be overridden at an individual subprogram level (see [RM 7.3.4](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-7-3-4.html)). It seems worth considering something like that for SPARK for private types.

[...]

PR #131, opened 2025-03-05
RFC for the Modifies aspect

clairedross:
Propose a RFC for #126

### DIFF
--- a/considered/rfc-spark-modifies-aspect.md
+++ b/considered/rfc-spark-modifies-aspect.md
@@ -25,23 +25,23 @@ Guide-level explanation
 =======================
 
 It is possible to annotate subprograms with side-effects with a ``Modifies``
-aspect. This aspect splits the input space in a number of disjoint cases, and,
-for each case, lists the parts of objects that are both inputs and outputs of
-the subprogram that are modified by the subprogram. Other objects are
-necessarily left unchanged.
+aspect. This aspect provides a list of clauses, each specifying a set of objects
+that are modified by the program, possibly with an additional guard introduced
+by the ``when`` keyword. Objects not mentioned in one of these clauses, or
+objects only mentioned in clauses whose guard evaluates to False, are
+necessarily left unchanged by the subprogram.
 
 This allows users to easily specify that some of their outputs are left
 unchanged on some paths. Here is an example. The global variable ``G`` is only
-written by ``Write_G_If_B`` if its parameter ``B`` evaluates to True. Its
-``Modifies`` aspect states that ``Write_G_If_B`` does not modify anything if
-``B`` is False:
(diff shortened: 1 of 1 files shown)


sttaft:
The proposal looks very good in general, but it might not be intuitively clear that these alternatives are exclusive, nor what it means when there is no "others".   Is it "others => all" or "others => null"?  I see that you are currently defaulting to "others => all" but that seems potentially misleading. I also wonder about "others" vs., say, "else".

More generally, I wonder whether a different syntax might make answers to these questions more obvious.

Brainstorming here for an alternative syntax... Perhaps you might turn things around, and say:

   Modifies => (object|object_list [when condition]{; ...}).

A given object or component can only be mentioned once, with unconditional object names coming first.  In this case, "others" would represent all objects not mentioned earlier.  If there is no others, only the components mentioned are ever modified.

Your examples above would then be:

```
   procedure Write_G_If_B (B : Boolean) with
     Global => (In_Out => G),
     Modifies => (not B => null);
```
becomes
```
   procedure Write_G_If_B (B : Boolean) with
     Global => (In_Out => G),
     Modifies => (others when B);
```
and
```
   procedure Write_G1_F1 (X : in out RR) with
     Modifies => (others => X.G1.F1);
```
becomes
```
   procedure Write_G1_F1 (X : in out RR) with
     Modifies => (X.G1.F1);
```
Other examples would be:
 `  Modifies (A; (B, C) when X > 5; D when Y < 2)
`
Clearly there are advantages and disadvantages of each syntax.  I suppose my main suggestion is to include many more examples, and perhaps mention alternative syntaxes, either ones that were rejected (and why), or ones that might be still worth consideration.  It is really only through examples that you will reach a point of confidence that the syntax "feels" right and intuitively communicates the desired semantics.

clairedross:
I cannot say that I have considered all possible other syntaxes, but the one you propose has a clear disadvantage (to me): it necessitate  to be able to make sure that each object is mentioned only once. However, this might actually depend on the value of inputs, if the object contains indexed components. An important advantage of the syntax I propose is that it is close to Contract_Cases, so SPARK users would be familiar with it.

clairedross:
I added your proposal in the alternatives. I must say I find it nice even if we cannot require that objects are never duplicated.

sttaft:
> I added your proposal in the alternatives. I must say I find it nice even if we cannot require that objects are never duplicated.

I suppose there is no particular reason to require uniqueness, presuming you simply "union" the set of objects/components identified by all those clauses whose conditions are True.

clairedross:
Yes, and then both alternatives are really equivalent.

clairedross:
I have added a separate commit to use this alternative instead.

PR #168, opened 2025-09-12
Importing unconstrained arrays from C or addresses

QuentinOchem:
Following improvements of so called fat-pointers that enable implementation of array slice access, we should now be able to import directly C arrays as unconstrained Ada arrays - or any contiguous piece of memory.

### DIFF
--- a/features/rfc-import-array-from-address.md
+++ b/features/rfc-import-array-from-address.md
@@ -1,7 +1,6 @@
 - Feature Name: Import Array from Address
 - Start Date: 2025-05-06
-- RFC PR: (leave this empty)
-- RFC Issue: (leave this empty)
+- Status: Proposed
 
 Summary
 =======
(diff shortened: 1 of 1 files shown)


tonunaks:
Thanks @QuentinOchem, I suggested some wording changes. Please check that I didn't alter the original intent.

jklmnn:
I know it already has been merged but it's still not clear to me what the improvement is over
```ada
type Arr is array (Integer range <>) of Integer;

function Create_C_Array_Of_Int (Size : Integer) return System.Address;
pragma Import (C, Create_C_Array_Of_Int, "create_c_array_of_int);

V : Arr (1 .. N) with
   Import,
   Convention => C,
   Address => Create_C_Array_Of_Int (N);
```

QuentinOchem:
> I know it already has been merged but it's still not clear to me what the improvement is over
> 
> ```ada
> type Arr is array (Integer range <>) of Integer;
> 
> function Create_C_Array_Of_Int (Size : Integer) return System.Address;
> pragma Import (C, Create_C_Array_Of_Int, "create_c_array_of_int);
> 
> V : Arr (1 .. N) with
>    Import,
>    Convention => C,
>    Address => Create_C_Array_Of_Int (N);
> ```

Hi @jklmnn ,

Thanks for the question :-)

This syntax is working in some situations, but does not allow you to represent a heap value in Ada. V is a variable on the stack, you can't take an access to it and have this access "escape" the scope of V.

Issue #159, opened 2025-08-29
[feature] OOP Attributes


tonunaks:
### Summary

TBD

### Motivation

TBD

### Caveats and alternatives

TBD

PR #175, opened 2025-10-24
Subprogram-valued Aspect Declarations

denismazzucato:
Part of #159

### DIFF
--- a/features/rfc-oop-attributes.rst
+++ b/features/rfc-oop-attributes.rst
@@ -5,14 +5,27 @@
 Summary
 =======
 
+This RFC introduces a shorthand notation to declare subprogram-valued aspects.
+
 Motivation
 ==========
+
+Currently, to specify a subprogram-valued aspect for a type, one needs to first
+declare a subprogram, that most of the times is never referred to elsewhere,
+and then link it to the type via an aspect specification or attribute
+definition clause. This pattern creates unnecessary boilerplate code,
+disconnects the aspect specification from the denoted subprogram declaration,
+and introduces potential name resolution issues.
 
 Guide-level explanation
 =======================
 
-Today, specifying a subprogram attribute of a type requires to declare a
-subprogram and then to use an aspect or an attribute to link it to a type, e.g.
+Subprogram-valued Aspect Declarations
+-------------------------------------
+
+Currently, specifying a subprogram-valued aspect for a type requires to first
+declare the subprogram and then apply the corresponding aspect specification or
+attribute definition clause. For example:
 
 .. code-block:: ada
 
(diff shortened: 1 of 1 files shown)


features/rfc-oop-attributes.rst
```diff
@@ -5,68 +5,382 @@
...
 
+Unfortunately, the direct attribute definitions inside a scoped record clashes
+with the class type rules where each subprogram is required to be primitive and
+following the rules of First_Controlling_Parameter (e.g., see ``T'Write`` above
+where ``Item`` is the second formal, not the first). The rule's motivation
+comes, for instance,
+from the fact that class primitives are called with the prefix notation, and
+that no primitive operation should be defined afterwards.
+
+To resolve this clash, we could relax the rules for class types to allow
```
QuentinOchem: We should only allow that for attributes at this stage. As they can't be called directly, there's no risk of confusion with other primitives.

features/rfc-oop-attributes.rst
```diff
@@ -5,68 +5,382 @@
...
+   type T is class record
+      procedure T'Constructor (Self : in out T);
       procedure T'Destructor (Self : in out T);
+
+      procedure T'Write
+        (Stream : not null access Ada.Streams.Root_Stream_Type'Class;
+         Item : in T);
    end record;
 
+Unfortunately, the direct attribute definitions inside a scoped record clashes
```
denismazzucato: Old terminology "direct attribute definitions". I should keep it consistent and use the one Tucker suggested: "subprogram-valued aspect declarations". I'll fix that.

ronan-d: Can you point me to the discussion where the new name appeared?

features/rfc-oop-attributes.rst
```diff
@@ -5,68 +5,382 @@
...
-      Stream : not null access Ada.Streams.Root_Stream_Type'Class;
+   procedure T'Write
+     (Stream : not null access Ada.Streams.Root_Stream_Type'Class;
       Item : in T);
 
-In addition, this can be used as a scoped attribute, possibly with re-ordering
-of parameters to ensure that the first parameter is conceptually the one
-"owning" the primitive, e.g.:
+Under the hood, the compiler expands this subprogram declaration into an
+*internal* subprogram declaration and corresponding aspect specification. The
```
ronan-d: > Under the hood, the compiler expands this subprogram declaration into an
> *internal* subprogram declaration and corresponding aspect specification

I think this should be rephrased to be more at the "language specification" level and less at the "compiler implementation" level. For example:

> This specifies the Write attribute of T to be an anonymous subprogram that has the profile and body of the procedure declaration.

However there is no occurrence of "anonymous subprogram" in the reference manual, even though there are many "anonymous type" and a few "anonymous object". So this can surely be improved.

features/rfc-oop-attributes.rst
```diff
@@ -5,68 +5,382 @@
...
+internal subprogram is not visible in the source code, and thus cannot be
+called directly. The name of the internal subprogram is determined solely by
+the aspect name, meaning that the internal subprogram denoted by
+``T'Write`` will have the same name of another declaration like ``U'Write``.
```
ronan-d: > The name of the internal subprogram is determined solely by
> the aspect name, meaning that the internal subprogram denoted by
> ``T'Write`` will have the same name of another declaration like ``U'Write``.

But using this feature for two different types in the same scope is expected to work, right?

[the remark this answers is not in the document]
ronan-d: > This is essential in type derivation since it permits overloadings,

Can you give an example?

> it may produce name resolution errors that we didn't anticipate and that the user may find counterintuitive since the name would be different in the source code but not in the compiled program.

But using the feature for two different types in the same scope is expected to work, right?

features/rfc-oop-attributes.rst
sttaft: I suggested "declaration" because unlike aspect specifications or attribute definitions, this is actually *declaring* a subprogram, and it requires an associated body to be provided.  So that makes it pretty different from any other "specification" or "definition".  Also, generally "specification" is part of a larger syntactic construct, whereas "declarations" tend to be a complete syntactic construct, ending with a ";".

features/rfc-oop-attributes.rst
```diff
@@ -5,68 +5,382 @@
...
+
+.. code-block:: ada
+
+   type Custom is limited private record;
+
+   procedure Custom'Put_Image
+     (Buffer : in out Ada.Strings.Text_Buffers.Root_Buffer_Type'Class;
+      Item : Custom);
+
+Here, a call to ``Custom'Put_Line (...)`` calls the aspect
```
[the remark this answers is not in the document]
denismazzucato: Yes my bad I kept writing `Put_Line` without realizing that actually I meant just the `Image` attribute... 

Just to clarify everything with the right name this time. When calling for instance:
```ada
My_Object : Custom := ...;
...
Ada.Text_IO.Put_Line(My_Object'Image);
```
the subprogram resolved by the attribute `'Image` is `Custom'Put_Image`.

The point I'm trying to make here is just that the attribute called and the one defined with declarations introduced in this rfc often coincides but not always. Does it make it more clear?

If you feel like this is good to have I'll fix the right attribute name, otherwise if it is more confusing than else I'll just remove it.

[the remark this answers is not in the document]
sttaft: Pointing out that 'Image, 'Wide_Image, and 'Wide_Wide_Image are all defined in terms of 'Put_Image is probably more of a distraction than helpful at this point.  Similarly 'Input is by default defined in terms of 'Read, but that seems somewhat irrelevant to this discussion.

On a somewhat separate topic, note that I suggested in a comment elsewhere that the equivalent aspect specification might best be thought of as something like:

    `with Put_Image => @'Put_Image`

meaning that the "nonoverridable" name is something like "`@'Put_Image`" rather than "`T'Put_Image`".

features/rfc-oop-attributes.rst
```diff
@@ -5,68 +5,382 @@
...
+the aspect name, meaning that the internal subprogram denoted by
+``T'Write`` will have the same name of another declaration like ``U'Write``.
+
+Reducing the Gap Between Aspect Call and Denoted Subprogram
+-----------------------------------------------------------
+
+For an Ada programmer, this shorthand helps connecting the aspect to its
+denoted subprogram. Even though an aspect call (e.g., ``T'Write (...)``) is not
+*strictly speaking*
+a subprogram call, under the hood the compiler transparently resolves it to the
```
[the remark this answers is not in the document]
denismazzucato: Will rephrase with this.

features/rfc-oop-attributes.rst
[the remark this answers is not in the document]
sttaft: > Thanks Ronan for the detailed example you presented above, I also agree with your analysis.
> 
> And to your suggestion Tucker, Instead of "@" is filled in with the associated type's name, I'd rephrase it as: in `@'Constant_Indexing` the placeholder `@` is filled with the name of the ancestor of the associated type (allowing itself as "first" ancestor here) which first specified `Constant_Indexing`. This modification is required to make `T'Constant_Indexing` and `T2'Constant_Indexing` share the same implicit name, but not with `Other'Constant_Indexing` to avoid name clashes with other unrelated entities.

Actually, I realize what I wrote was ambiguous.  What I meant is that the name used in the equivalent aspect specification is literally "`@'<aspect_name>`" but when you write the subprogram-valued declaration, the programmer replaces the "@" with the particular type name for which it is being declared.  Furthermore, when one of these subprograms is inherited, the @ is replaced with the new (derived) type's name in the implicit declaration, similar to the way the formal parameter types are replaced.

> 
> I don't know if this is going against Ronan's idea of not having to use an equivalence rule in terms of an implicit name (in this case the implicit name is the special `@'Constant_Indexing` here). But I think with this use of `@` we could give a more intuitive and less case-by-case definition for this proposal, it would be simply the equivalence with its extension in terms of `@` and then we define what `@` is replaced by. What do you think?

Yes, that is what I was trying to suggest.

features/rfc-oop-attributes.rst
```diff
@@ -5,68 +5,357 @@
 Summary
 =======
 
+This RFC introduces a shorthand notation to declare subprogram-valued aspects.
+
 Motivation
 ==========
 
+Currently, to specify a subprogram-valued aspect for a type, one needs to first
+declare a subprogram, that most of the times is never referred elsewhere, and
```
ronan-d: Typo: "referred elsewhere" -> "referred to elsewhere".

features/rfc-oop-attributes.rst
```diff
@@ -5,68 +5,357 @@
...
+   type Cursor is new Positive;
+
+   function Container'First (C : Container) return Cursor;
+   function Container'Next (C : Container; Pos : Cursor) return Cursor;
+   function Container'Has_Element (C : Container; Pos : Cursor) return Boolean;
+   function Container'Element (C : Container; Pos : Cursor) return Integer;
+
+All the subprogram-valued aspect
+declarations of ``Container`` are primitive operations.
+The usual rules of primitive operations apply.
```
ronan-d: Can you give an second example, for the nonprimitive case? (The one already here is a translation of the GNAT-specific `Iterable` aspect I think, and it does require the subprograms to be primitive operations).

denismazzucato: This would occur for all aspects that require a subprogram as value and that subprogram needs not to have a profile that will make them primitive. Currently, we can find such examples with class-wide aspects, such as:

```ada
   type T is tagged null record;
   
   --  nonprimitive
   procedure T'Class'Write
     (Stream : not null access Ada.Streams.Root_Stream_Type'Class;
      Item   : T'Class) is null;
```

But in the future more may be added, and I couldn't find anything in the RM that requires a subprogram used as aspect value that needs to be primitive.

features/rfc-oop-attributes.rst
denismazzucato: Fixed.

PR #181, opened 2025-11-13
Tweak example in assigments RFC

ronan-d:


ronan-d:
I put a rationale for the change in the commit message, but I just realized that GitHub's UI makes it easy to miss.

QuentinOchem:
> I put a rationale for the change in the commit message, but I just realized that GitHub's UI makes it easy to miss.

Interesting - so indeed we're trying to show two different things. Would you mind if I put your example in the rationale bit, so that we keep the description of cleanup in the general case?

PR #185, opened 2025-12-10
Composing by-constructor and non-by constructor types

QuentinOchem:


features/rfc-oop-constructors.rst
```diff
@@ -732,6 +732,87 @@ An object value in a constructor is considered initialized once the `Super` and
...
+      type Non_Tagged is private;
+      procedure Non_Tagged'Constructor (Self : in out Non_Tagged) is abstract;
+
+This does require to describe how constructors are expressed on non-tagged
+types which is not part of this proposal.
+
+Note that the above also says that allocation of such object without a copy
+or aggregate is illegal. E.g.:
+
+. code-block:: ada
```
denismazzucato: Missing a dot.

features/rfc-oop-constructors.rst
```diff
@@ -732,6 +732,87 @@ An object value in a constructor is considered initialized once the `Super` and
 `Initialze` aspects have been computed. Formally, the role of the constructor
 is to establish further properties than the initialization.
 
+Composing by-constructors types with non by-constructor types
+-------------------------------------------------------------
+
+By-constructor types can be included in types that are not build by
```
denismazzucato: What is the exact definition of a "by-constructor type"? a type declared via a tagged record that has explicitly declared a constructor?

denismazzucato (changes requested):
Minor comments below.

One thing that needs to be assessed, is that once a by-constructor type is component of a type `T`, then `T` becomes "implicitly" a by-constructor type. This is essential because then an implicitly by-constructor type might be used as component of other types, and in turn influence their kind of initialization. Also I think the same rules should apply to parent types, not only to components, no?

As a side note, I think we need to clear the implicit/explicit constructors. There are several differences that I didn't get when first reading the rfc, for instance that default initialization is permitted if a type only have implicit constructors, thus that type is not per se "by-constructor" but still we might be able to build it with both `'Make` parameterless and default initialization. Or again that a regular type could be an implicit by-constructor, and as a consequence to forbid default initialization. I just wanted to point out that this difference isn't clear in the rfc but I'm not asking for any edit about it now.

QuentinOchem:
> Minor comments below.
> 
> One thing that needs to be assessed, is that once a by-constructor type is component of a type `T`, then `T` becomes "implicitly" a by-constructor type. This is essential because then an implicitly by-constructor type might be used as component of other types, and in turn influence their kind of initialization. Also I think the same rules should apply to parent types, not only to components, no?

There's indeed an issue of composition here. So we could take two stances - we can indeed implicitly  propagate the "by constructor" aspect of types. This has impact on what you can do with types in particular with regards to default initialization - you should not be able to have a controlled be by constructor. The alternative is to only propagate whether or not a type has default initialization. E.g. in :

```ada
type X is tagged record
   null;
end record;

procedure X'Constructor (Self : in out X; V : Integer);

type C is record
   F: X;
end record;
--  C is not by constructor, but doesn't have default initialization

type C2 is record
   F2 : C;
end record
--  C2 is not by constructor, but doesn't have default initialization either
```

I believe that as far as limited goes (ie types without copy constructors), this propagation is already there.

> 
> As a side note, I think we need to clear the implicit/explicit constructors. There are several differences that I didn't get when first reading the rfc, for instance that default initialization is permitted if a type only have implicit constructors, thus that type is not per se "by-constructor" but still we might be able to build it with both `'Make` parameterless and default initialization. Or again that a regular type could be an implicit by-constructor, and as a consequence to forbid default initialization. I just wanted to point out that this difference isn't clear in the rfc but I'm not asking for any edit about it now.

Ok - requires clarification indeed - I'll work on that. There's a difference between tagged and class types here:
- only tagged types that have constructors are by constructors. So if there's no constructor specified (and you don't derive from a type already by constructor), you're not a by-constructor type and there's no implicit constructor. I would argue that these types can't be build by 'Make (this is more for clarity than a real technical requirement).
- all class record types are by constructor whether or not they have an explicit one.

Does that make sense?

features/rfc-oop-constructors.rst
```diff
@@ -732,6 +732,87 @@ An object value in a constructor is considered initialized once the `Super` and
...
+Further extensions of the language may allow to provide more specification to
+the private view to allow this - we may want to be able to write:
+
+.. code-block:: ada
+
+      type Non_Tagged is private;
+      procedure Non_Tagged'Constructor (Self : in out Non_Tagged) is abstract;
+
+This does require to describe how constructors are expressed on non-tagged
+types which is not part of this proposal.
```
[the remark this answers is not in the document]
QuentinOchem: Indeed - this is a leftover from an initial version of the proposal. Untagged record needs to be handled separately. I'll clean.

features/rfc-oop-constructors.rst
QuentinOchem: Will fix.

features/rfc-oop-constructors.rst
QuentinOchem: Yes - or that is derived from a by-constructor type. I'll make that explicit.

QuentinOchem:
> A few more comments I've been thinking about (I can't comment on lines that are not modified by this PR but I think these are relevant):
> 
> * this section should be before the section about copy constructor and should define what is a by-constructor type so that it is clear when used afterwards
> * above it's written that: "if not specified, a default copy constructor is automatically generated" I'd rephrase it to "for by-constructor types, if not specified, an implicit copy constructor is automatically generated". Otherwise, any type will have a copy constructor and thus will be by-constructor and will create problems later when another type extends or derives from it

Agreed - will fix.

features/rfc-oop-constructors.rst
```diff
@@ -732,6 +732,87 @@ An object value in a constructor is considered initialized once the `Super` and
...
+
+   type Non_Tagged is record
+      X : Root;
+   end record;
+
+   type Arr is array (Integer range <>) of Root;
+
+In this case, the rules are as follows:
+
+- [1] If the by-constructor type provides both a by-copy and a parameterless
```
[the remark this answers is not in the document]
QuentinOchem: Sounds good.

tonunaks (approved):

PR #190, opened 2025-12-17
[DRAFT] constructors for non-tagged objects

QuentinOchem:


features/rfc-oop-non_tagged_lifecycle.rst
```diff
@@ -0,0 +1,613 @@
...
+
+Aggregates
+----------
+
+Aggregates on non-tagged types behave exactly as described in the OOP
+aggregates and assignments RFC for tagged types: an aggregate is not a way to
+build an object behind the back of its constructors. It is a constructor call
+followed by modifications of the resulting object. The type must therefore
+provide a parameterless constructor visible at the point of the aggregate, and
+the aggregate is equivalent to 'Make followed by component updates.
```
MartinClochardPro, line 367: What if there are discriminants ? This cannot be done by component updates, and the default-value fields cannot be provided by the default constructor. Furthermore, that rule effectively forces the copy constructor to call the default constructor on an internal temporary whose state will be immediately erased (since aggregates for each variants are the only way to make the copy correctly).

MartinClochardPro: In the same vein, how is this supposed to interact with limited and build-in-place ? You cannot copy limited component, however initializing them by a build-in-place aggregate is currently possible.

PR #199, opened 2026-02-04
alternate API to the formatter

QuentinOchem:


denismazzucato:
Just a thought, why not to have also something like "Open_Flat_Record" or "Open_Unnested_Record" that would not format the beginning and end of a record, so to avoid having unreadable records from deep record hierarchies (I think Steven mentioned this problem at some point). In that case the first step is called via `Open_Record` but then proceeds via the flattened one recursively for the parents.

features/rfc-to_string.md
swbaird: Ordinary_Fixed_Point_Type (with a comma).
And four uses below (in subtype ranges).
Alternatively,  just list how this package spec differs from the Flare version.

swbaird: What is intended here (and in the next 6 declarations)? Subtype declarations (so "type" should be "subtype") or derived type declarations (so the word "new" is missing)? I'm guessing subtypes.  Either way, some fix is needed.

swbaird: ";" needed
function must be abstract because result type is abstract

features/rfc-to_string.md
```diff
@@ -129,26 +138,251 @@ package Flare.Strings.Text_Formatters is
...
 
-   type Root_Formatter_Parameters is abstract class record
-      --  Provided for future extensions - should contain various formating
-      --  parameters necessary for default formating.
+      procedure Put_Address (
+         Self      : in out Root_Formatter_Type;
+         Buffer    : in out Root_Buffer_Type;
+         Type_Name : UTF_Encoding.UTF_8_String;
+         Item      : System.Address;
+         Format    : Root_Formatter_Parameters;
```
swbaird: extra ";"
Ditto for the next 25 subprograms.

a subprogram cannot be a primitive operation of two tagged types (Root_Formatter_Type and Root_Formatter_Parameters).
Ditto for the next 25 subprograms.

features/rfc-to_string.md
```diff
@@ -184,33 +418,15 @@ package Flare.Strings.Text_Formatters is
...
-         Buffer        : in out Root_Buffer_Type;
-         Name          : UTF_Encoding.UTF_8_String := "";
-         Is_Constraint : Boolean;
-         Format        : Root_Formatter_Parameters
-      );
-
       procedure Close (
-         Self   : in out Root_Buffer_Type;
+         Self   : in out Root_Formatter_Type;
          Buffer : in out Root_Buffer_Type;
```
swbaird: Extra ";"

features/rfc-to_string.md
```diff
@@ -184,33 +418,15 @@ package Flare.Strings.Text_Formatters is
...
       );
    end Root_Formatter_Type with private;
```
swbaird, line 458: looks like incomplete editing after copying the Flare version. Perhaps replace this line and the next 3 with
   type Default_Formatter_Type is new Root_Formatter_Type with private;
?

features/rfc-to_string.md
swbaird: missing "constant" (declaration is illegal without it)

features/rfc-to_string.md
```diff
@@ -129,26 +138,251 @@ package Flare.Strings.Text_Formatters is
...
+      ) is abstract;
+
+      --  Array Support  --
+
+      procedure Open_Array (
+         Self           : in out Root_Formatter_Type;
+         Buffer         : in out Root_Buffer_Type;
+         Type_Name      : UTF_Encoding.UTF_8_String;
+         Format         : Root_Formatter_Parameters;
+      );
```
swbaird: Should this (and the following dozen or so procedures) be abstract?

swbaird (commented):
A few minor cleanup issues, although the problem with a procedures illegally being primitives of two different tagged types (Formatter and Formatter_Parameters) might require adding a nested package or some similar minor reorganization.

I wrote these comments with the Ada Compatibility Mode version in mind, although some of them may also apply to the Flare version.

features/rfc-to_string.md
QuentinOchem: Will fix.

Indeed for the primitive issue, this should be Root_Formatter_Parameters'Class

features/rfc-to_string.md
QuentinOchem: Yes - will fix.

features/rfc-to_string.md
QuentinOchem: Will fix

features/rfc-to_string.md
QuentinOchem: This is indeed incomplete editing. Will fix.

features/rfc-to_string.md
QuentinOchem: I don't understand this one - why can't you create variable of local types?

features/rfc-to_string.md
QuentinOchem: ... actually no - this is a class type, to in this case Root_Formatter_Parameters is not controlling / primitive (only the first parameter is). But this is an issue for the Ada case.

features/rfc-to_string.md
QuentinOchem: Ada spec need to be cleaned, will do.

QuentinOchem: Indeed, will do.

QuentinOchem: Will fix.

swbaird: Good point. The change I suggested would be needed if Default_Formatter_Type is declared with a private extension instead of a visible implementation-defined set of components. With that change (which may actually be what we want) it would be illegal to declare a variable of a not-yet-completed private extension type, but a deferred constant declaration would be ok [and similarly if any of those implementation-defined component types turn out to be not-yet-completed private types]. But that's not what the RFC says, so I agree that declaring a variable object is ok. I'll leave this one unresolved so that you can decide if we want  Default_Formatter_Type to be declared as a private extension (possibly with implementation-defined primitive subprograms). If you want to leave things "as is" then feel free to resolve this one.

QuentinOchem: Interesting. We may want to rediscuss that at some point - maybe we'll need to create separate packages to allow variables of private types if we get there.

QuentinOchem:
> Just a thought, why not to have also something like "Open_Flat_Record" or "Open_Unnested_Record" that would not format the beginning and end of a record, so to avoid having unreadable records from deep record hierarchies (I think Steven mentioned this problem at some point). In that case the first step is called via `Open_Record` but then proceeds via the flattened one recursively for the parents.

Missed that one, apologies. I think the theory is that the formatter can delay the output to detect various kind of records it cares about - Flat / Nested being just two special cases. Unless I'm misreading your comment, that should be doable today.

features/rfc-to_string.md
swbaird: If we don't really need a variable, then a  deferred constant seems like the simplest solution.

Otherwise, one approach is a nested package for the type, as in
    package P1 is
        package Q is
             type T is private;
        private
            ...;
        end Q;
        Obj : T;
    end P1;
but that reorganization seems heavy.

A deferred constant of a reference type might be a lighter approach:
    package P2 is
         type T is private;
         type T_Ref (Ref : not null access T) is limited null record with Implicit_Dereference => Ref;
         Obj  : constant T_Ref;
    private
          ...;
    end P2;

QuentinOchem: At this stage, and given that we have another object for parameters, deferred constant is probably good indeed. I'll make the change.

QuentinOchem: As a matter of fact, you would not want anyone to change the default object, so constant is most definitely indeed the right pattern.

features/rfc-to_string.md
```diff
@@ -539,130 +745,339 @@ end Some_Task_Type'To_String;
...
-   type Integer_Type is Discrete_Type range Signed_Type .. Modular_Type;
-   type Real_Type is Scalar_Type range Float_Type .. Ordinary_Type;
-   Type Fixed_Type is Scalar_Type range Decimal_Type .. Ordinary_Type;
-   type Composite_Type is Ada_Type range Array_Type .. Task_Type;
-
    type Root_Formatter_Parameters is abstract tagged record
-      --  Provided for future extensions - should contain various formating
-      --  parameters necessary for default formating.
+      --  Provided for future extensions - should contain various formatting
+      --  parameters necessary for default formatting.
```
swbaird, line 789: Should this be an abstract tagged private type?

features/rfc-to_string.md
```diff
@@ -539,130 +745,339 @@ end Some_Task_Type'To_String;
...
 
       null;
    end record;
 
    function Create
       (Self   : in out Root_Formatter_Parameters;
-       Format : String) return Root_Formatter_Parameters
+       Format : String) return Root_Formatter_Parameters is abstract;
 
    type Root_Formatter_Type is abstract tagged record
```
swbaird: Should this be an abstract tagged private type?

features/rfc-to_string.md
```diff
@@ -539,130 +745,339 @@ end Some_Task_Type'To_String;
...
+Ada.Strings.Text_Formatters will be introduced with the necessary tagged types
+and default implementations.
 
-The packages provided for Ada will look as follow:
+The package provided for Ada will look as follows:
 
 ```ada
 package Ada.Strings.Text_Formatters is
 
-   type Ada_Type is
```
swbaird, line 554: Is the declaration of type Ada_Type intended to be gone?
Do we need at least something like "type Ada_Type is (... same enumeration as Flare version ...);" 
Ditto for the subtypes defining ranges of this type.

features/rfc-to_string.md
```diff
@@ -116,121 +112,335 @@ formatter to structure the output, as opposed to the indent primitives. E.g.:
...
+         Item      : Ada.Numerics.Big_Numbers.Big_Reals.Big_Real;
+         Format    : Root_Formatter_Parameters)
+      is abstract;
+
+      procedure Put_Decimal_Type (
+         Self          : in out Root_Formatter_Type;
+         Buffer        : in out Root_Buffer_Type;
+         Type_Name     : UTF_Encoding.UTF_8_String;
+         Integer_Value : Interfaces.Unsigned_64;
+         Decimal_Value : Interfaces.Unsigned_64;
```
swbaird: What is the Decimal_Value parameter? I think we need an example of what actual parameter should be generated in a call to Put_Decimal_Type. Perhaps this was meant to somehow represent T'Small.

features/rfc-to_string.md
```diff
@@ -116,121 +112,335 @@ formatter to structure the output, as opposed to the indent primitives. E.g.:
...
+         Format        : Root_Formatter_Parameters)
+      is abstract;
+
+      procedure Put_Ordinary_Fixed_Point_Type (
+         Self          : in out Root_Formatter_Type;
+         Buffer        : in out Root_Buffer_Type;
+         Type_Name     : UTF_Encoding.UTF_8_String;
+         Integer_Value : Interfaces.Unsigned_64;
+         Decimal_Value : Interfaces.Unsigned_64;
```
swbaird: Same questions as above for Put_Decimal_Type .

swbaird (commented):
A couple of minor questions about fixed point images.

features/rfc-to_string.md
```diff
@@ -539,130 +745,339 @@ end Some_Task_Type'To_String;
...
+
+   procedure Close
+      (Self   : in out Root_Formatter_Type;
+       Buffer : in out Root_Buffer_Type'Class)
+   is abstract;
+
+   type Default_Formatter_Type is new Root_Formatter_Type with private;
 
-   Default_Formatter : Default_Formatter_Type;
+   Default_Formatter : constant Default_Formatter_Type;
```
swbaird: If we change the type of this deferred constant to Root_Formatter_Type'Class, then there is no need to make the declaration of Default_Formatter_Type visible.
Would that be a desirable change?

features/rfc-to_string.md
QuentinOchem: In my mental model, if you have 10.59, the Integer value is 10 and the decimal is 59. This is probably the easiest for the formatter to then implement. Does that work>

features/rfc-to_string.md
QuentinOchem: I think there's no other use for it with the new design - ie we now have specific subprograms to differentiate as opposed to one subprogram takin the kind as parameter.

features/rfc-to_string.md
QuentinOchem: Tagged private works - probably cleaner indeed. Will fix.

features/rfc-to_string.md
QuentinOchem: Yes - we have Default_Formatter_Type anyway for default (and of course that's consistent with the decision to have these subprograms abstract).

features/rfc-to_string.md
QuentinOchem: I'm not sure about the impact - but I didn't realize that constant 'Class was even a possibility in this situation. Arguably cleaner indeed, will fix.

features/rfc-to_string.md
swbaird: It is not clear to me that we need different interfaces for the two different kinds of  fixed point types, but I have no objection to making that distinction. I think that in both cases we want to pass in the integer representation of the the fixed point value and some representation of the 'Small for the fixed point type, probably as a pair of integers representing a numerator/denominator pair. [When I say "integer", I just mean some appropriate integer type, not specifically Standard.Integer .]

features/rfc-to_string.md
swbaird: You usually can't declare an object of an indefinite subtype (such as T'Class) without providing an initial value, but it is ok for the initial declaration of a deferred constant.

features/rfc-to_string.md
QuentinOchem: Maybe we don't need two interfaces indeed - I would be equally happy with just one.

So you're suggesting 3 parameters? (Integer_Value, Decimal_Value, Small)? What type do you want for Small?

swbaird: A different 3 parameters: The integer representation of the value, 'Small numerator, and 'Small denominator.
So if we have a fixed point type with Small = 2.0 / 3.0, then the mathematical value 6.0 would generate parameters 9 (the integer representation of 6.0), 2 ('Small numerator), and 3 ('Small denominator).

swbaird: We can have two parameters of type Interfaces.Unsigned_64 to represent the
'Small value. Perhaps named Numerator_Of_Small and Denominator_Of_Small?
I don't like the name Small_Numerator because it is confusing; but perhaps
Scale_Numerator and Scale_Denominator?

We can also have a comment or precondition or something to indicate that those
two parameter values are each nonzero and that they are relatively prime
(so the fraction is expressed in lowest terms).

But what type to use for the Value parameter, the integer representation of
the value being output?

We solved this problem for integer types by having multiple Put_Integer_Xxx
procedures. We could do the same thing here with 4 new procedures:
   Put_Fixed_Point_8/16/32/64.
Or would could just provide one procedure with a parameter that is "big
enough" (but if we are going to do that, then why not also do it in the
integer case?).

swbaird: If we do go with the 'one procedure with a parameter that is "big enough"' approach for integers too, I think we would still need different procedures for the signed and unsigned case. If we tried to combine them, what type would we use for the Value parameter?

QuentinOchem:
@tonunaks I think we're ready for a merge on this now - @swbaird and I have resolved the last issues.

features/rfc-to_string.md
```diff
@@ -266,22 +266,54 @@ package Flare.Strings.Text_Formatters is
...
-         Type_Name     : UTF_Encoding.UTF_8_String;
-         Integer_Value : Interfaces.Unsigned_64;
-         Decimal_Value : Interfaces.Unsigned_64;
-         Format        : Root_Formatter_Parameters)
+      procedure Put_Fixed_Point_8 (
+         Self              : in out Root_Formatter_Type;
+         Buffer            : in out Root_Buffer_Type;
+         Type_Name         : UTF_Encoding.UTF_8_String;
+         Value             : Interfaces.Integer_8;
+         Scope_Numerator   : Interfaces.Integer_8;
```
swbaird: Scope_Numerator => Scale_Numerator, right?
Similar change in other places.

swbaird (commented):
One comment (but it applies in several cases).

features/rfc-to_string.md
QuentinOchem: Rah - Absolutely. Will fix.

swbaird (approved):

PR #201, opened 2026-02-12
New RFC spark-constant-in-iterator

### LINKED ISSUE #191: [feature] Use Contant_Reference instead of Element in Iterable
### Summary

The Iterable aspect is a GNAT specific aspect that provides user defined iteration over containers. Currently it requires `First`, `Next`, and `Has_Element` functions to iterate over the cursors of the container (at least), and an Element function can be added to obtain iteration over the content of the container (`for Item of C loop`, instead of `for Position in Container` loop). I would be interesting to allow providing a `Constant_Reference` function returning an anonymous access-to-constant view of the element instead of the `Element` function.

As an example, we currently have:
```ada
   type List (Capacity : Count_Type) is private
   with
     Iterable                  =>
       (First       => First,
        Next        => Next,
        Has_Element => Has_Element,
        Element     => Element),

   function Element (Container : List; Position : Cursor) return Element_Type;
```

We could instead allow:
```ada
   type List (Capacity : Count_Type) is private
   with
     Iterable                  =>
       (First       => First,
        Next        => Next,
        Has_Element => Has_Element,
        Constant_Reference => Constant_Reference),

   function Constant_Reference (Container : List; Position : Cursor) return not null access constant Element_Type;
```

A loop:
```ada
for E of Container loop
  P (E);
end loop;
```

Would be equivalent to:
```ada
declare
   Position : Cursor := First (Container);
begin
   while Has_Element (Container, Position) loop
      declare
         Ref : access constant Element_Type := Constant_Reference (Container, Position);
         E : Element_Type renames Ref.all;
      begin
         P (E);
      end;
   end loop;
end;
```

### Motivation

The use of Constant_Reference would avoid copying the element, in particular in quantified expressions where using a while loop is not an alternative (at least for spark).

### Caveats and alternatives

A small caveat is that, since E is a renaming and not an object, it would not be possible to use it at certain places, for example inside Global and Depends contracts. It sound minor to me though.

As a follow-up, we could consider having a way to supply a Reference function that would allow direct mutation inside the container. It is notably more complicated to do that in a SPARK-compatible way though, because of aliasing restrictions, so it may be better to keep it as a separate RFC.

tonunaks:
https://github.com/AdaCore/ada-spark-rfcs/issues/191

### DIFF
--- a/features/rfc-spark-constant-in-iterator.rst
+++ b/features/rfc-spark-constant-in-iterator.rst
@@ -1,0 +1,138 @@
+- Feature ID: spark-constant-in-iterator
+- Start Date: 2026-01-09
+- Status: Design
+
+Summary
+=======
+
+The Iterable aspect is a GNAT specific aspect that provides user defined iteration over containers. Currently, it requires `First`, `Next`, and `Has_Element` functions to iterate over the cursors of the container at minimum. An `Element` function can be added to obtain iteration over the content of the container (`for Item of C loop`, instead of `for Position in Container` loop). In a use-case when the elements of the container are 
+fetched only for reading the data, this method is unnecessarily slow as it
+involves copying the element. 
+The performance issue can be solved by adding an optional `Constant_Reference` function returning an anonymous access-to-constant view of the element instead of copy.
+
+Motivation
+==========
+
+The use of Constant_Reference would avoid copying the element, in particular in quantified expressions where using a while loop is not an alternative (at least not for SPARK).
+
+Guide-level explanation
+=======================
+
+
+As an example, we currently have:
+
+```ada
+   type List (Capacity : Count_Type) is private
+   with
+     Iterable  =>
+       (First       => First,
+        Next        => Next,
+        Has_Element => Has_Element,
+        Element     => Element),
+
+   function Element (Container : List; Position : Cursor) return Element_Type;
+```
+
+The current proposal adds an alternative prototype:
+
+```ada
+   type List (Capacity : Count_Type) is private
+   with
+     Iterable =>
+       (First       => First,
+        Next        => Next,
+        Has_Element => Has_Element,
+        Constant_Reference => Constant_Reference),
+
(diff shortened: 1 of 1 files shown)


features/rfc-spark-constant-in-iterator.rst
```diff
@@ -0,0 +1,126 @@
...
+Rationale and alternatives
+==========================
+
+An alternative to this proposal would be [5.5.3 Procedural Iterators](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-5-5-3.html). That would solve the
+problem of potentially dangling reference. Reference would be passed as IN parameter and it's lifetime is strictly limited. However, it would create a different kind of performance issue as the compiler optimisation would be 
+disabled in some cases. The LD group the decided to drop this alternative
+because, despite of being safer, wouldn't solve the original problem outlined
+of the summary of this RFC.
+
+Another possible enhancement would be allowing private functions in the
```
sttaft, line 91: Can you clarify what you mean by a "private function", perhaps with an example?

tonunaks: @Nikokrock, do you have an example for this? You mentioned that this has been discussed earlier.

Nikokrock:
```ada

type Bytes is private
   with Iterable => (First       => First,
                     Next        => Next,
                     Has_Element => Has_Element,
                     Element     => Unsafe_Get);
```

If you want to have an efficient implementation, it's better to provide a Get_Element that does not checks bounds has this check is already done by Has_Element. Would be great to make in that case Unsafe_Get private (and so only usable through the iterator)

clairedross: I am not sure I understand what you mean by "bounds".  Note that Get_Element might have a precondition (not executed in production) and not necessarily defensive coding.

clairedross (approved):
I don't understand the private functions comments in the Rationale and alternatives and in the Future possibilities sections, but I don't think it is blocking.

Issue #210, opened 2026-04-21
Is Default_Dispatching_Calls a configuration pragma?


ronan-d:
The RFC for alternative dispatching rules has:

> A new library / partition level pragma is introduced, Default_Dispatching_Calls.

I understand that "library / partition level pragma" implies that it's a partition-wide configuration pragma.

But a bit later, the RFC has the following pseudocode:

> ```
> package A is
>
>    pragma Default_Dispatching_Calls (Off);
> 
>    type Root is tagged null record;
> ...
> ```

which suggests that you can turn the pragma on and off at a finer level of granularity.

I'm fairly confident that there's an ambiguity in the RFC that needs to be resolved,
but I'm not sure which interpretation to keep.

ronan-d:
@QuentinOchem, what do you think?

QuentinOchem:
@ronan-d indeed, the intent was Library _level_ and configuration pragma. That is, you can activate / deactivate for a given library unit, but not at a finer granularity. If that makes sense, I'll update the RFC accordingly.

ronan-d:
Great, please do update the RFC. ~Note that there was a nice RM addition in Amendment 1 to Ada 2022: the definition of *context pragmas* in [2.8](http://www.ada-auth.org/standards/22aarm_w_amd1/html/AA-2-8.html) (7/6).~

~I suggest using "context pragma" for the RFC, that corresponds better to the intent than "configuration pragma" that strictly speaking can only be specified on the main library unit~

Updated, see Tuck's comment below.

sttaft:
> I suggest using "context pragma" for the RFC, that corresponds better to the intent than "configuration pragma" that strictly speaking can only be specified on the main library unit.

I am not sure I agree, since a context pragma can *only* appear in the place where a "use" clause might be allowed, and is generally intended to be something that qualifies a "with" clause somehow, e.g. like a "pragma Elaborate".  A configuration pragma can be used to specify an aspect that applies to the set of compilation units in a given source file.  There is also the now obsolescent notion of a "library unit pragma" -- see J.15(7/5).  In general, we now encourage the use of library-unit-level aspect specifications instead of library unit pragmas.

ronan-d:
I see, thanks for the clarification.

ronan-d:
Fixed by #210.

PR #213, opened 2026-04-29
refined granularity of pragma (for #210)

QuentinOchem:


features/rfc-oop-dispatching.rst
```diff
@@ -14,7 +14,10 @@ Guide-level explanation
 Dispatching and Class Wide Views
 --------------------------------
 
-A new library / partition level pragma is introduced, Default_Dispatching_Calls.
+A new library-level / partition level pragma is introduced, Default_Dispatching_Calls.
+A given library unit can either be on or off - it is not possible to adjust the
+granuarity at a finer level.
```
ronan-d: Typo: "granuarity" -> "granularity".

features/rfc-oop-dispatching.rst
```diff
@@ -14,7 +14,10 @@ Guide-level explanation
 Dispatching and Class Wide Views
 --------------------------------
 
-A new library / partition level pragma is introduced, Default_Dispatching_Calls.
+A new library-level / partition level pragma is introduced, Default_Dispatching_Calls.
+A given library unit can either be on or off - it is not possible to adjust the
```
ronan-d: Is it really "library unit" and not "compilation unit"? This matters if, for example, there can be a spec where Default_Dispatching_Calls is Off while it is On in the corresponding body.

ronan-d (changes requested):

features/rfc-oop-dispatching.rst
QuentinOchem: Will fix.

features/rfc-oop-dispatching.rst
QuentinOchem: Indeed - compilation unit - will fix.

ronan-d (approved):

PR #217, opened 2026-05-07
Class fixes

QuentinOchem:


features/rfc-oop-primitives.rst
```diff
@@ -61,27 +62,51 @@ bodies. The following demonstrates the above:
...
+   end C;
+
+The body section of a class type follows the same rule as a package body: it
+is mandatory if and only if at least one primitive declared in the spec
+requires completion. If every primitive is fully defined in the spec (for
+example, all of them are expression functions, or there are none), the
+``type body <name> is ... end <name>;`` section must be omitted.
+
+Primitives declared within a type can only be called via prefix notation. With
+class types, primitives can only be declared in the scope of the type.
```
marc-adacore: I find the wording confusing.

> Primitives declared withing a type

As primitives are required to be declared within a type, I would skip this extra qualification and the following sentence.

```
Class primitives are always called via prefix notation
```

features/rfc-oop-primitives.rst
```diff
@@ -36,22 +37,22 @@ bodies. The following demonstrates the above:
          F : Integer;
 
          procedure Prim (Self : in out R; V : Integer);
-       end record
+       end R
```
marc-adacore, line 40: Does that mean that this RFC now also have https://github.com/AdaCore/ada-spark-rfcs/blob/master/features/rfc-mandatory_end_designator.md as a dependency?

features/rfc-oop-primitives.rst
QuentinOchem: Sounds good - will fix.

features/rfc-oop-primitives.rst
QuentinOchem: We can decide to implement that as a dependency or not - what's clear is that since class record is a new construction, there's no need to support a legacy form. So maybe that doesn't need to be a dependency?

tonunaks (dismissed):

tonunaks (dismissed):

PR #218, opened 2026-05-11
Mark unchecked-union-discriminant ready for prototyping

tonunaks:
Add an example from the issue

### DIFF
--- a/features/rfc-spark-ghost-unchecked-union-discriminant.md
+++ b/features/rfc-spark-ghost-unchecked-union-discriminant.md
@@ -1,6 +1,6 @@
 - Feature Name: spark_ghost_unchecked_union_discriminant
 - Start Date: 2026-04-02
-- Status: Design
+- Status: Ready for prototyping
 
 Summary
 =======
@@ -11,8 +11,8 @@ cases for verification purposes.
 Motivation
 ===========
 
-
-Ada mandates that discriminant of unchecked union types (used to interface C union) cannot be read, as per legality rule B.3.3 (9/5). This is
+Ada mandates that discriminant of unchecked union types (used to interface C
+union) cannot be read, as per legality rule B.3.3 (9/5). This is
 cumbersome in the context of verification with SPARK, as the discriminant
 carry information pertinent to the specification. For example, some
 subprograms may only work for a subclass of discriminants. Without the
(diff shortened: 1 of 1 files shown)


features/rfc-spark-ghost-unchecked-union-discriminant.md
```diff
@@ -30,6 +30,35 @@ Legality rule B.3.3 (9/5) is relaxed: the list of cases in which a
...
+   end record with
+     Unchecked_Union;
+   
+   function Read_F_Int (X : My_Rec) return Integer with
+     Pre => (Static => X.Is_Int);
+   
+   function Read_F_Int (X : My_Rec) return Integer is (X.F_Int);
+   
+   procedure Write_F_Int_If_B (X : in out My_Rec; B : Boolean) with
+     Post => (Static => (if B then X = (True, 12) else X = X'Old));
```
MartinClochardPro, line 60: The RFC text as-is does not make that equality check legal, which is arguably an oversight. Should we extend the RFC to also allow the extra cases that Ada forbids because they do implicit discriminant access ? I extracted the cases in question from the RM, I am in favor of adding at least the first two:

- predefined equality operator when any of the operator is an unconstrained unchecked_union type
- membership test to constrained unchecked_union type

The following case feels fishy, because I am not sure what is even supposed to happen with primitives under a lost-content derivation:
- Conversion from derived unconstrained unchecked_union type to base unconstrained which does not have unchecked_union.

Finally, there are also cases for Read/Write predefined attributes, but as far as I can tell, this should never be useful in non-executable code.

sttaft: I agree that equality and membership could be well defined in ghost code.  I'd have to see some examples of conversion in ghost code to make an intelligent decision there.

tonunaks: Good that it came up here! We should not rush to merge incomplete text. Let's take care of the clarifications first. Since we already started the discussion here, the most straightforward thing for you would be to check out the same patch and put your commit on top of mine.

MartinClochardPro: I do not have the permissions to push, so I nested a PR.

PR #223, opened 2026-05-14
RFC: Model_of aspect

### LINKED ISSUE #197: [feature] Model_Of aspect to specify external subprograms
### Summary

This RFC proposes a new Ada aspect, `Model_Of`, intended for use by analysis tools to associate additional specification information with an existing subprogram. The aspect allows tools to link a *model* subprogram, declared separately in a specification package, to an original subprogram, and to treat any aspects applied to the model as if they were applied to the modeled subprogram for analysis purposes.

Suppose an existing package contains a subprogram that cannot be modified:

```ada
package Math is
   function Sqrt (X : Float) return Float;
end Math;
```

An analysis tool may require a more robust specification. A separate specification package can be written:

```ada
package Math_Models is
   function Sqrt (X : Float) return Float
     with
       Model_Of => Math.Sqrt,
       Pre  => X >= 0.0,
       Post => Sqrt'Result >= 0.0;
end Math_Models;
```

### Motivation

Analysis tools such as formal verifiers, abstract interpreters, or linters often require additional semantic information that cannot always be directly embedded in production source code. Reasons include:

* The original subprogram belongs to third-party or legacy code.
* The original code must remain free of tool-specific annotations.
* Alternative or abstract specifications are needed for verification or modeling purposes.

Currently, there is no standardized mechanism in Ada to express that one subprogram declaration *models* another, nor to propagate aspects from a model declaration to the original subprogram in a tool-recognizable way.

The `Model_Of` aspect addresses this gap by enabling the declaration of a subprogram that mimics an existing one and explicitly links to it. Analysis tools can then use this relationship to augment or replace the original specification without modifying the original source.

### Caveats and alternatives

* Introduces a concept that is meaningful only to analysis tools.
* Requires tools to implement additional consistency checks.
* May lead to fragmented specifications if overused.

However, not introducing this feature forces tools to rely on naming conventions, external configuration files, or source rewriting.

tonunaks:


features/model_of.md
```diff
@@ -0,0 +1,464 @@
...
+can still participate in analysis.
+
+### (3) Toolchain integration
+
+Integrating the model decorator step with the rest of the toolchain is outside
+the scope of this proposal.
+
+### (4) Code annotations in the decorated unit
+
+Special syntax for code addotations is activated by `pragma Source_Annotations (On)`.
```
marc-adacore:
```suggestion
Special syntax for code annotations is activated by `pragma Source_Annotations (On)`.
```

features/model_of.md
```diff
@@ -0,0 +1,464 @@
...
+
+(3) A workflow for integrating the model step with the rest of the
+toolchain.
+
+(4) Syntax for code annotations in the **decorated unit**.
+
+(5) Compiler modifications to take code annotations into account when displaying
+diagnostics messages.
+
+Step (3) is outside the scope of this proposal as it depends on the implementation choiches on step (2).
```
sttaft: "choices" not "choiches"

features/model_of.md
```diff
@@ -0,0 +1,464 @@
...
+step inside the compiler. 
+
+Motivation
+==========
+
+It may not be always possible or reasonable to include a formal contract in the specification of the subprograms. Some examples of scenarios, where formal contracts in source code may not be feasible:
+
+* When analysing legacy code, one may want to avoid unnecessary changes in the original sources.
+
+* Formal contracs may be long and complex. Mixing them with functional code may reduce the readability and maintainability of both.
```
sttaft: "contract" rather than "contracs"

features/model_of.md
```diff
@@ -0,0 +1,464 @@
...
+
+### Validity rules for sidecar units
+
+Only SPARK aspects that have no effect on execution may be specified in the
+model — for example `Pre`, `Post`, `Global`, `Depends`, `Contract_Cases`,
+`Test_Case`, `Subprogram_Variant`, `Ghost`, and `Annotate` — and only provided
+that the same aspect doesn't exist in the actual subprogram spec yet.
+Overriding or changing an existing aspect is not permitted.
+
+Annotation can be applied to generinc instance, but not to generic.
```
sttaft: "generic" rather than "generinc", and perhaps add articles: "... to a generic instance, but not to a generic"

features/model_of.md
tonunaks: Thanks! Fixed.

features/model_of.md
tonunaks: Thanks! Fixed.

features/model_of.md
tonunaks: Thanks!

features/model_of.md
tonunaks: Fixed

tonunaks:
@MartinClochardPro, please check if the validity rule on line 153 is now appropriate? I rephrased it to reference `Assertion_Policy` now

features/model_of.md
```diff
@@ -0,0 +1,488 @@
...
+
+Reference-level explanation
+===========================
+
+### (1) Sidecar unit grammar
+
+```
+decoration_unit ::=
+ "model" "package" package_name "is"
+ { decorated_subprogram_declaration }
```
thvnx, line 321: The grammar only admits `decorated_subprogram_declaration`, is subprogram-only the intended final scope? SPARK contracts also operates on packages/types/globals. Worth a line in the RFC on whether that's in or out.

features/model_of.md
```diff
@@ -0,0 +1,488 @@
...
+the analysis tools instead of the original main unit.
+
+`pragma Source_Annotations (On)` instructs the compiler to apply special processing
+to annotations representing code locations.
+
+A given subprogram may carry contracts in at most one of the main unit and the
+sidecar unit. If both files supply a pre- or postcondition for the same
+subprogram, the model decorator tool reports an error indicating the conflicting
+line in each file. This rule prevents silent shadowing of production contracts
+by sidecar contracts.
```
thvnx: I think this contradict what's stated at lines 157-159.

features/model_of.md
```diff
@@ -0,0 +1,488 @@
...
+```ada
+pragma Source_Annotations (On)
+
+--  my_math_decorated.ads  (decorated specification, consumed by analysis tools)
+--# @Source_Reference@ 1 "my_math.ads" 1
+package My_Math is
+ function Sqrt (X : Float) return Float
+--# @Source_Reference@ 3 "my_math_models.ads" 1
+ with
+--# @Source_Reference@ 5 "my_math_models.ads"
```
thvnx, line 194: Shouldn't line number be 4 here (and therefore no marker at all)?

features/model_of.md
```diff
@@ -0,0 +1,488 @@
...
+ Pre  => X >= 0.0,
+ Post => Sqrt'Result >= 0.0;
+--# @Source_Reference@ 3 "my_math.ads" 2
+ function Log (X : Float) return Float;
+end My_Math;
```
thvnx, line 199: Is it expected that the private part of my_math.ads is missing from this unit?

features/model_of.md
```diff
@@ -0,0 +1,488 @@
...
+`pragma Source_Annotations (On)` instructs the compiler to apply special processing
+to annotations representing code locations.
+
+A given subprogram may carry contracts in at most one of the main unit and the
+sidecar unit. If both files supply a pre- or postcondition for the same
+subprogram, the model decorator tool reports an error indicating the conflicting
+line in each file. This rule prevents silent shadowing of production contracts
+by sidecar contracts.
+
+If a sidecar unit exists for a package without a corresponding main source
```
thvnx, line 253: This feels counter intuitive: the feature is meant to stay "read-only" on the sources I guess, so creating a unit here seems odd. Why not make it a hard error instead? Reconstructing a spec from its contracts also looks non-trivial?

features/model_of.md
```diff
@@ -0,0 +1,488 @@
...
+A companion sidecar body may provide contracts for `Pow` instead of `Sqrt`
+and define a helper function in the model package:
+
+```ada
+--  my_math_models.adb  (sidecar body, not compiled by the regular build)
+
+--  additional included packages
+with Helper_Package; use Helper_Package;
+
+model package My_Math is
```
thvnx, line 137: The `model` keyword distinguishes decorated declarations from the ones copied through verbatim, but that could be an aspect instead of new vocabulary:

```ada
function Pow (X, Y : Float) return Float
   with Model,  Pre  => Greater_Than (X, 0.0), Post => Greater_Than (Pow'Result, 0.0);
```

This reuses Ada's existing aspect grammar (no new keyword, no new gramme rule,…), and would greatly simplify tooling: libadalang could parse sidecar units with its normal grammar, and LKQL rules and IDE support (navigation, completion, highlighting) would work out of the box.

It would also read better when decorating an entity that is itself named Model:

```ada
model procedure Model with …
-- vs.
procedure Model with Model, …
```

features/model_of.md
```diff
@@ -0,0 +1,488 @@
...
+ return Pow (X, 0.5);
+ end Sqrt;
+end My_Math;
+```
+
+A companion sidecar body may provide contracts for `Pow` instead of `Sqrt`
+and define a helper function in the model package:
+
+```ada
+--  my_math_models.adb  (sidecar body, not compiled by the regular build)
```
thvnx, line 132: The RFC doesn't define a naming convention for the model unit files. Could it either define one or state explicitly that naming is unconstrained? This also brings gprbuild support into play: how sidecar units are linked to their Ada specifications, and how they are kept out of the build.

I would propose <unit>.adm: it keeps the same base name as the spec/body, and makes clear that it isn't Ada code. gprbuild could likely handle that easily, extending the existing naming attributes accordingly.

features/model_of.md
```diff
@@ -0,0 +1,488 @@
...
+ "end" package_name ";"
+
+decorated_subprogram_declaration ::=
+ "model" subprogram_specification
+ "with" decoration_aspect_list ";"
+
+decoration_aspect_list ::=
+ decoration_aspect { "," decoration_aspect }
+
+decoration_aspect ::=
```
thvnx, line 331: I think the original issue for that feature wasn't limited to SPARK aspects: https://github.com/AdaCore/ada-spark-rfcs/issues/197. In particular, the GNAT SAS team is interested in taint aspects. The rule should be generalized to all aspects that have no effect on execution, not just the SPARK ones.

features/model_of.md
tonunaks: Yes. To keep the scope manageable, we decided to restrict the proposal to subprogram contracts only. I added this statement to the summary.

features/model_of.md
tonunaks: Good catch! Updated.

features/model_of.md
tonunaks: The reason for this was to support adding ghost code without limiting it to expression functions.
I added a sentence explaining this in the RFC. Does that make the rule less weird?

features/model_of.md
tonunaks: Indeed, SPARK here was a leftover. Removed.

features/model_of.md
tonunaks: The first versions had a proposal for a naming scheme. That was removed later by the group, arguing that RFC should deal with the language definition. Organization into files and the usage of different tools are out of scope here.

features/model_of.md
thvnx: Got it, thanks!

features/model_of.md
thvnx: Understood, thanks!

features/model_of.md
tonunaks: Originally, we had something in that vein:

```
package models My_Math is
   function Sqrt (X : Float) return Float
     with
       Model_Of => My_Math.Sqrt,
       Pre  => X >= 0.0,
       Post => Sqrt'Result >= 0.0;
end My_Math;
```

Then the aspect was dropped in favor of specific syntax. It seemed redundant with `package models` and too verbous. What you are proposing is much more concise. With the changes we have made in the meantime (e.g. addin non-model code that should be copied as is), it might make more sense. Let's discuss on todays meeting

features/rfc-model.md
```diff
@@ -0,0 +1,488 @@
...
+`pragma Source_Annotations (On)` instructs the compiler to apply special processing
+to annotations representing code locations.
+
+A given subprogram may carry contracts in at most one of the main unit and the
+sidecar unit. If both files supply a pre- or postcondition for the same
+subprogram, the model decorator tool reports an error indicating the conflicting
+line in each file. This rule prevents silent shadowing of production contracts
+by sidecar contracts.
+
+If a sidecar unit exists for a package without a corresponding main source
```
thvnx: OK, starting to understand. That means that such a unit can't have model declarations, only regular (ghost) code. That's clearer but I would add that "model entries are not permitted in that case" IIUC.

features/rfc-model.md
```diff
@@ -0,0 +1,488 @@
...
+package My_Math is
+ function Sqrt (X : Float) return Float
+--# @Source_Reference@ 3 "my_math_models.ads" 1
+ with
+--# @Source_Reference@ 5 "my_math_models.ads"
+ Pre  => X >= 0.0,
+ Post => Sqrt'Result >= 0.0;
+--# @Source_Reference@ 3 "my_math.ads" 2
+ function Log (X : Float) return Float;
+end My_Math;
```
tonunaks: No, the private part was added later, and I forgot to update the synthesized file. Done now.

features/rfc-model.md
tonunaks: Updated. I also added corresponding validity rule.

features/model_of.md
tonunaks: This is now reorganized, according to the agreement on the last LD meeting

features/model_of.md
```diff
@@ -0,0 +1,577 @@
+- Feature Name: model_of
+- Start Date: 2026-04-02
+- Status: Design
+
+Summary
+=======
+
+This RFC proposes a mechanism for attaching analysis-oriented contracts,
+consisting of aspects that do not affect execution with assertions
+disbled (such as `Pre`, `Post`, `Global`, and `Test_Case`), to subprograms
```
thvnx: disbled -> disabled

features/model_of.md
```diff
@@ -0,0 +1,577 @@
...
+Annotation can be applied to a generic instance, but not to a generic.
+
+The fragments of the sidecar unit that do not carry the `Model` aspect
+and are thus copied as-is shall have a valid Ada syntax. The model decorator
+step does not check them, assuming that, in the event of conflicts with other
+elements of the program, the compilation step will discover them.
+
+An annotation can be applied to a child unit using the unit's qualified name.
+
+The decoration step shall stop with an error when it cannot find a suprogram
```
thvnx: suprogram -> subprogram

features/model_of.md
```diff
@@ -0,0 +1,577 @@
...
+The model decorator step reads both the main unit and the sidecar unit, and produces the decorated unit.
+
+For specifications, the step merges the package specification with the sidecar model declarations.
+It injects contracts into the corresponding source declarations.
+For bodies, the step performs the same merge on the body: it reads the actual package body and the sidecar body, resolves declarations carrying `Model` against body specifications, and injects the remaining aspects into the matched declarations.
+
+In both cases, fragments of the sidecar that do not carry the `Model` aspect are copied through unchanged, and the decorated output preserves origin information using linemarkers. The `Model` aspect itself is never emitted.
+
+```ada
+pragma Source_Annotations (On)
```
thvnx: missing semicolon

features/model_of.md
```diff
@@ -0,0 +1,577 @@
...
+the run of lines that started at `my_math.ads` line 3, so their origin is
+already implied by the preceding marker.
+
+For a body, the decorated output is produced in the same way.
+The model contracts are merged into the actual body declaration, and the
+origin linemarkers preserve their relationship to both the original body and
+the sidecar body.
+
+```ada
+pragma Source_Annotations (On)
```
thvnx: missing semicolon

features/model_of.md
```diff
@@ -0,0 +1,577 @@
+- Feature Name: model_of
```
thvnx: I guess the feature should be renamed to `model` instead of `model_of`.

thvnx:
Only a few minor comments left. LGTM otherwise!

features/model_of.md
tonunaks: I am fine either way. Since we have now renamed the aspect, it makes sense. However, I do not see a problem keeping the feature name as proposed originally and using a different pragma in the text. I added this to tomorrow's meeting notes.

tonunaks: Renamed, as per the decision of the LD meeting.

PR #225, opened 2026-05-18
Update Default_Dispatching_calls RFC

### LINKED ISSUE #216: oop-dispatching: complete controlling result example
The example in [the paragraph about controlling results](https://github.com/AdaCore/ada-spark-rfcs/blob/master/features/rfc-oop-dispatching.rst#dispatching-on-returned-types) is only partial.

Here's my attempt at completing it:

```ada
pragma Default_Dispatching_Calls (On);

procedure Main is
   package P is
      type T is tagged null record;
      function Make return T; -- primitive
   end P;

   package body P is
      function Make return T is
      begin
         return (null record);
      end Make;
   end P;

   use P;

   Obj1 : T'Class := T'(null record);
   Obj2 : T'Class := Make; -- illegal
begin
   Obj1 := Make; -- legal; use Obj1'Tag to dispatch
end Main;
```

But the line flagged with `-- illegal` is currently accepted by GNAT, so I don't think this correctly illustrates the intent of the RFC.

@QuentinOchem, what do you think?

ronan-d:
Closes #216.

### DIFF
--- a/features/rfc-oop-dispatching.rst
+++ b/features/rfc-oop-dispatching.rst
@@ -134,22 +134,96 @@ For example:
 Note that this is a problem when integrating with current Ada, pedantic Flare
 does not support multi-parameter dispatching.
 
-Dispatching on Returned Types
------------------------------
+Dispatching assigments
+----------------------
 
-A tag indeterminate dispatching call is illegal (as it is the case today). For
-example:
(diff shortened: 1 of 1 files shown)


ronan-d:
CC @swbaird

QuentinOchem (approved):

PR #230, opened 2026-06-16
Aggregates fixes

QuentinOchem:


features/rfc-oop-aggregates-and-assignments.rst
```diff
@@ -197,27 +124,18 @@ needs to be maintained equal to the parents.
...
-     (Self : in out Child; From : Root'Class; Copied_As : Tag) is
-   begin
-      if not Is_Descendant_At_Same_Level (Copied_As, Child'Tag) then
-         --  The assignment was made through a view above Child, so B was not
-         --  copied and may be inconsistent with A. Re-derive it. Note that this
-         --  is decided from the assignment view (Copied_As), not from From's
-         --  type: Root (C1) := Root (C2) is partial even when From is a Child.
+      Self'Super'ASsign (From);
+      Free (Self.B);
+      Self.B := new Integer'(From.B.all);
```
ronan-d, line 132: Is `From.B` legal here? IIUC `From` is of type `Root.Class`.

QuentinOchem: Indeed - will add an else to the if and add a downcast there.

tonunaks (approved):

PR #233, opened 2026-07-08
Refine pragma flare RFC

viljar-indus:
Add more explicit examples and refine the use of pragma flare

### DIFF
--- a/features/rfc-flare.md
+++ b/features/rfc-flare.md
@@ -1,4 +1,4 @@
-- Feature Name:flare-pragma
+- Feature Name: flare-pragma
 - Start Date: 2026-02-26
 - Status: Ready for prototyping
 
@@ -10,16 +10,17 @@
 
 ## Flare Versioning
 
-The Ada/Flare language is activated at the unit level (spec and body). Flare
-can be turned on at the child package level, once that's the case all further
-children must also be Ada/Flare. Flare packages may depend on Ada units, the
-reverse is not possible. Furthermore, if a specific Flare version is specified,
-children units or dependents must all be at least of that specific version, or
-further.
+The Flare language is activated at the compilation unit level (spec or
+body). A unit to which no Flare pragma applies is *regular Ada*; a unit in Flare
+mode is either *pedantic Flare* (pragma `Flare`) or *extensions* mode (pragma
+`Flare_Extensions`). Flare mode can be turned on at the child package level.
+Once that's the case all further children must also be in Flare mode. Flare
+packages may depend on regular Ada units, the reverse is not possible.
 
-Flare versions implement semantic numbering, Major.Minor. Beside the minor
-versions under the major 0 which is considered beta, minor versions are not
-allowed to introduce backward-incompatible changes.
+Flare allows for specific versioning. Flare versions implement semantic
+numbering, Major.Minor. Beside the minor versions under the major 0 which is
+considered beta, minor versions are not allowed to introduce
+backward-incompatible changes.
 
 The version of Flare described in this document is Flare 0.1.
 
@@ -28,8 +29,15 @@ may decide on whether or not a unit is Flare, and what version. This may be
 a default, a flag, a file extension or any other indication that said tool
 may use.
 
-A unit can be marked Flare introducing the pragma Flare either on its
-specification, or implementation if there's no specification. E.g.:
+If a specific Flare version is specified, child and dependent units must all
+be at least of that specific version, or further. A body, by contrast, must
+match the version of its specification exactly (see [Pragma
+Scope](#pragma-scope)).
+
+### Pragma Flare
+
+A unit can be marked Flare introducing a configuration pragma Flare either on
+its specification, or implementation. E.g.:
 
 ```ada
 pragma Flare;
@@ -39,7 +47,43 @@ package P is
 end P;
 ```
 
-When provided without arguments, the tool is instructed to select the most
+### Pragma Scope
+
+The pragma Flare only applies to the compilation unit it is placed on. Like
+other configuration pragmas placed before a library level package
+specification, it is not propagated to the corresponding package body. The
+body must therefore repeat the pragma to be Flare as well.
+
+```ada
+-- p.ads
+pragma Flare;
+package P is
+   --  This is Flare code
+end P;
+
+-- p.adb
+package body P is -- ERROR: Spec is in Flare mode whereas the body is not
+   --  This is regular Ada code
+end P;
+```
+
+The reverse is not allowed either: a body may not be Flare unless its spec is
+also Flare. A spec and its body must have the same Flare status.
+
+```ada
+-- p.ads
+package P is
+   --  This is regular Ada code
+end P;
+
+-- p.adb
+pragma Flare;
+package body P is -- ERROR: Body is Flare whereas the spec is not
+   --  This is Flare code
+end P;
+```
+
+When provided without a version, the tool is instructed to select the most
 recent version available. A developer may also provide a version with specific
 version-aware flare pragmas, e.g.:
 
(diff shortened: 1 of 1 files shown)


features/rfc-flare.md
```diff
@@ -28,18 +27,65 @@ may decide on whether or not a unit is Flare, and what version. This may be
 a default, a flag, a file extension or any other indication that said tool
 may use.
 
-A unit can be marked Flare introducing the pragma Flare either on its
-specification, or implementation if there's no specification. E.g.:
+If a specific Flare version is specified, children unit or dependents must all
```
tonunaks: child and dependent units

features/rfc-flare.md
```diff
@@ -28,18 +27,65 @@ may decide on whether or not a unit is Flare, and what version. This may be
...
+
+-- p.adb
+package body P is -- ERROR: Spec is in Flare mode whereas the body is not
+   --  This is regular Ada code
+end P;
+```
+
+In the future, the pragma could be made "sticky" so that a pragma Flare on the
+spec also applies to the corresponding body, removing the need to repeat it.
+See [Future possibilities](#future-possibilities).
```
tonunaks: I would remove this from here and leave only under future-possibilities. Apart from making the spec cleaner, it also makes the next paragraph unambiguous. Currently, it is not immediately clear where "The reverse" refers.

features/rfc-flare.md
```diff
@@ -28,18 +27,65 @@ may decide on whether or not a unit is Flare, and what version. This may be
...
 
-A unit can be marked Flare introducing the pragma Flare either on its
-specification, or implementation if there's no specification. E.g.:
+If a specific Flare version is specified, children unit or dependents must all
+be at least of that specific version, or further.
+
+### Pragma Flare
+
+A unit can be marked Flare introducing a configuration pragma Flare either on
+its specification, or implementation if there's no specification. E.g.:
```
tonunaks: this contradicts with the statement on line 70, where you say that the pragma can be put in body even if it doesn't exist in the spec. This statement suggests that pragma Flare in body is accepted only when there is no specificaiton.

features/rfc-flare.md
```diff
@@ -70,6 +218,12 @@ package P is
 end P;
 ```
 
+Note that the use of pragma Flare_Extensions is close to GNAT's
+[`pragma Extensions_Allowed`](#gnat-pragma-extensions_allowed): both accept
+Ada semantics and add capabilities on top, the difference being that
+Flare_Extensions is versioned and limited to non-backward-incompatible Flare
```
tonunaks: Double negaition. Rather use "backward-compatible"

features/rfc-flare.md
```diff
@@ -165,6 +319,63 @@ introduce backward-incompatible changes in the pedantic Flare version.
...
+[experimental extensions][gnat-experimental-extensions]. It corresponds to the
+`-gnatX` and `-gnatX0` command-line switches, over which it takes precedence.
+
+Many of the features slated for Flare 0.1 are already available through this
+mechanism, including for example the conditional `when` constructs, the
+`continue` and `finally` statements, fixed lower bounds, constructors and
+destructors, structural generic instantiation, inference of dependent types,
+the `External_Initialization` aspect and string interpolation.
+
+This makes `Extensions_Allowed` very close in spirit to the
```
tonunaks: It is not clear from here whether `pragma Extensions_Allowed(All_Extensions)` includes Flare extensions if we decided to omit those from -gnatX0 in the future. Assuming it is not I'd rephrase this whole paragraph as follows:

This makes `pragma Extensions_Allowed` very close to the `pragma Flare_Extensions`. Both accept standard Ada and layer additional features on top, rather than defining a new language subset.  `Extensions_Allowed(All_Extensions)` activates curated and experimental GNAT-specific extensions. `Flare_Extensions` adds additional Flare features in their backward-compatible form. Unlike `Extensions_Allowed`, `Flare_Extensions` allows to specify the version of the Flare language.

features/rfc-flare.md
```diff
@@ -165,6 +319,63 @@ introduce backward-incompatible changes in the pedantic Flare version.
...
+[`pragma Flare_Extensions`](#pragma-flare_extensions): both accept standard Ada
+semantics and layer additional capabilities on top, rather than defining a
+strict new language. The main differences are that `Flare_Extensions` is
+versioned and only admits non-backward-incompatible Flare capabilities, so its
+feature set and semantics are pinned by the requested version, and that a
+`Flare_Extensions` unit must still satisfy the Flare dependency and child rules.
+
+The pedantic `pragma Flare`, by contrast, has no direct GNAT equivalent: it
+defines a specific, versioned subset of the language and rejects code outside
```
tonunaks: "language. It rejects"

Shorter sentences make it easier to parse

tonunaks (commented):
I added some grammar-oriented nits, and a few recommendations to make the text clearer.

Looks good, overall!

features/rfc-flare.md
viljar-indus: Fixed

features/rfc-flare.md
viljar-indus: Fixed

features/rfc-flare.md
viljar-indus: Fixed.

features/rfc-flare.md
```diff
@@ -28,18 +27,65 @@ may decide on whether or not a unit is Flare, and what version. This may be
...
+```
+
+In the future, the pragma could be made "sticky" so that a pragma Flare on the
+spec also applies to the corresponding body, removing the need to repeat it.
+See [Future possibilities](#future-possibilities).
+
+The reverse, however, is allowed: a body may be Flare even when its spec is
+not. If the part being exposed in the spec doesn't violate any restrictions,
+then the user of the module doesn't care whether the restriction was applied
+or not.
```
viljar-indus: When it comes to only specifying the body with flare does this mean that the unit is still an Ada Unit?

This is important when it comes to rules about child and dependent units?

features/rfc-flare.md
viljar-indus: Fixed.

features/rfc-flare.md
viljar-indus: Fixed

features/rfc-flare.md
viljar-indus: Fixed

viljar-indus (commented):
Thank you for your comments.

features/rfc-flare.md
tonunaks: As discussed live, let's drop this requirement (at least for now). As you explained, it is calling for trouble, and the value is not clear.

features/rfc-flare.md
```diff
@@ -10,16 +10,17 @@
...
-can be turned on at the child package level, once that's the case all further
-children must also be Ada/Flare. Flare packages may depend on Ada units, the
-reverse is not possible. Furthermore, if a specific Flare version is specified,
-children units or dependents must all be at least of that specific version, or
-further.
+The Flare language is activated at the compilation unit level (spec or
+body). A unit to which no Flare pragma applies is *regular Ada*; a unit in Flare
+mode is either *pedantic Flare* (pragma `Flare`) or *extensions* mode (pragma
+`Flare_Extensions`). Flare mode can be turned on at the child package level,
+once that's the case all further children must also be in Flare mode. Flare
```
tonunaks: level.
Once that's the case, all

features/rfc-flare.md
```diff
@@ -168,12 +335,280 @@ introduce backward-incompatible changes in the pedantic Flare version.
...
+compilation unit to which no such pragma applies is *regular Ada*; a unit to
+which a `Flare` pragma applies is in *pedantic Flare* mode; a unit to which a
+`Flare_Extensions` pragma applies is in *extensions* mode. The pedantic and
+extensions modes are together referred to as *Flare mode*.
+
+### Legality Rules — the version
+
+A `version` denotes a Major.Minor pair with the following semantics:
+
+* If a `minor_version` is omitted, the pragma denotes a version whose major
```
tonunaks: Given that we already have a rule to pick the most recent version without the version number, it seems natural to do the same for a missing minor version. It would be confusing to apply a different rule here.
I suggest to remove the open question and update this paragraph

viljar-indus: Fixed.

features/rfc-flare.md
```diff
@@ -70,16 +226,29 @@ package P is
...
+the extensions mode as "non-pedantic Ada Flare"; the latter is the
+`Flare_Extensions` mode described here.
+
+Note that the use of pragma Flare_Extensions is close to GNAT's
+[`pragma Extensions_Allowed`](#gnat-pragma-extensions_allowed): both accept
+Ada semantics and add capabilities on top, the difference being that
+Flare_Extensions is versioned and limited to backward-compatible Flare
+features.
+
+The current Flare versions are based of a subset of the Ada 2022 language.
```
thvnx: Not directly part of this PR but could you also list the Ada 2022 features that are not included in that subset?

tonunaks: I agree that this is out of scope for this RFC. The list of implemented features is already in GNAT RM. Everything outside this list is unimplemented. I would change the sentence above as

"Flare language is based on the Ada 2022 language subset supported by GNAT (consult the GNAT Reference Manual for the list of supported Ada 2022 features)."

viljar-indus: I am not sure what Quentin meant here.

Whether its the subset of Ada 2022 features that gnat has implemented or it relates to the features that become invalid in Flare, like the Ada style short circuit operators etc.

tonunaks: I am pretty sure, that here we are referring to the ada 2022 subset that is implemented by gnat

tonunaks (approved):

PR #234, opened 2026-07-09
Add paragraph about implicit references to unchecked_union discriminant

MartinClochardPro:


features/rfc-spark-ghost-unchecked-union-discriminant.md
```diff
@@ -30,6 +30,14 @@ Legality rule B.3.3 (9/5) is relaxed: the list of cases in which a
 name denoting a discriminant of a type with Unchecked_Union is allowed to
 occur is extended by ghost code with an assertion level depending on Static.
 
+Dynamic semantics B.3.3 (22/2) is modified by only raising Program_Error
+for the listed cases if they occur outside of ghost code with an
+assertion level depending on Static. (This cannot affect the dynamic
+semantics of any Ada program. The intent is to clarify that tools using
```
sttaft: An alternative approach would be to say that an unchecked union object is considered to have inferable discriminants anywhere in ghost code with an assertion level of Static.

MartinClochardPro: This is not equivalent, as your proposition would not allow default Read/Write. Although maybe we do not want them there ? I have no strong opinion on that topic. As far as I can tell, this has zero effects on any useful annotations, unlike the other three.

sttaft: I had not noticed the paragraphs about 'Read/'Write/'Input/'Output, and I agree that those would not be allowed by simply assuming the discriminants are inferable.  On the other hand, it seems odd for ghost code to be making any mention of these streaming attributes, so it might be a "feature" to not allow them for unchecked unions, even in static ghost code.

MartinClochardPro: Fine by me. I used inferable discriminant instead and added the consideration to 'rationale and alternatives'.

tonunaks: If I understand it correctly, then this is resolved. I'll merge the pull request.

PR #237, opened 2026-07-28
Propose to rename `At` attribute to `At_Label`

thvnx:
`At` collides with the Ada reserved word `at`. A reserved word cannot be used as an attribute designator without a special grammar case, so the bare `At` spelling is not straightforward to parse and support. `At_Label` avoids the collision (keeping the syntax unambiguous and easier for tools to adopt) and mirrors `Loop_Entry`, making clear it belongs to the same family of attributes that capture a value at an earlier point.

Just a proposal, feel free to drop it if it's too late in prototyping, but renaming `At` to `At_Label` would sidestep the reserved-word at collision and ease tooling support.

### DIFF
--- a/features/rfc-spark-values-at-labels.md
+++ b/features/rfc-spark-values-at-labels.md
@@ -7,10 +7,12 @@ Status: Ready for prototyping
 Summary
 =======
 
-This RFC introduces a new attribute, called ``At`` that can be applied to
+This RFC introduces a new attribute, called ``At_Label`` that can be applied to
 an expression to refer to a copy of its value saved at a preceding label.
 This attribute can be thought as a generalization of the ``Old`` and
-``Loop_Entry`` attributes.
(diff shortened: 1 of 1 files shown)


MartinClochardPro:
This is a discussion we already had in the meetings, although I cannot dig back in which from the notes. The usage was judged appropriate and hardly different in grammatical nature from the preexisting `'Range` for arrays. As for the argument about `'Loop_Entry`, the same argument can be made than `'At` mirror `'Old`. Finally, `'At` is more user-friendly due to concision, so there is a trade-off between easier for tools/easier for users. I do not feel like the argument presented here makes a sufficient case to revisit the decision at this point, especially since the feature is already implemented in several places.

thvnx:
Fair enough, the `'Range`/`'Old` precedents settle it and the extra handling on the tooling side, while not free, is manageable, so I'm happy to keep `'At`. Thanks @MartinClochardPro for the feedback.

Issue #240, opened 2026-07-31
[feature] SPARK_Mode_Enabled


kanigsson:
### Summary

Alternative to pragma/aspect SPARK_Mode, that accepts any static Boolean expression instead of "On" or "Off". Suggested name: SPARK_Mode_Enabled.

Example
```
package Configuration is
   Prove_Lib : constant := True;
end Configuration;

generic
   ...
package Lib with SPARK_Mode_Enabled => True is

   ...

private
   pragma SPARK_Mode_Enabled(Configuration.Prove_Lib);
   ...
end Lib;

package body Lib with SPARK_Mode_Enabled => Configuration.Prove_Lib is
   ...
end Lib;
```

Other than the name, and the fact that the pragma/aspect accepts Booleans, the semantics and placement rules would be identical to `SPARK_Mode`.

### Motivation

Clients of SPARKlib, and more generally clients SPARK libraries with generics, generally don't want to prove the code of the generic. But from SPARK point of view the code is part of the unit where it is instantiated, so it will be proved by gnatprove.

The commonly used solution is to move the private part and package body of the generic package (or the body of a generic subprogram) behind `SPARK_Mode => Off`.

But developers of the library in question want to prove the generic using a sufficiently general instantiation. This proof needs the body of the package (Lib in the example) to be under `SPARK_Mode => On`. With the `SPARK_Mode` pragma, this conflict can only be resolved with conditional compilation that duplicates a lot of code, or preprocessing.

### Caveats and alternatives

The main alternative we considered was the extension of `SPARK_Mode` to not only accept `On` and `Off`, but also static Booleans. We rejected this as those names could also refer to static booleans and this could be ambiguous.

PR #241, opened 2026-07-31
SPARK_Mode with Static Booleans RFC

kanigsson:
For #240

### DIFF
--- a/features/rfc-spark_mode-static-boolean.md
+++ b/features/rfc-spark_mode-static-boolean.md
@@ -1,0 +1,149 @@
+- Feature Name: SPARK_Mode with static Boolean expressions
+- Start Date: 2026-07-31
+- Status: Design
+
+Summary
+=======
+
+This RFC aims to extend the `SPARK_Mode` aspect and pragma so that, in addition
+to the identifiers `On`, `Off` and `Auto`, it accepts static Boolean
+expressions. Placement, inheritance and consistency rules are unchanged.
+
+Motivation
+==========
+
+Clients of SPARKlib, and more generally clients of SPARK libraries with
+generics, usually don't want to prove the code of the generic. But from the
+SPARK point of view, the code is part of the unit where it is instantiated, so
+it will be proved by gnatprove.
+
+The commonly used solution is to move the private part and package body of the
+generic package (or the body of a generic subprogram) behind SPARK_Mode => Off.
+
+But developers of the library in question want to prove the generic using a
+sufficiently general instantiation. This proof needs the body of the package
+(`Lib` in the example) to be under SPARK_Mode => On. With the SPARK_Mode pragma
+as it stands today, this conflict can only be resolved with conditional
+compilation that duplicates a lot of code, or preprocessing.
+
+Guide-level explanation
+=======================
+
+Wherever `SPARK_Mode => On` or `SPARK_Mode => Off` is allowed, one can also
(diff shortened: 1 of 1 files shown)


JoseRuizAdaCore:
I'd prefer extending `SPARK_Mode` to accept static Boolean expressions alongside `On` and `Off` literals, rather than adding a second aspect and pragma.

Regarding the ambiguity concern (within the Rationale and alternatives section), the ARM already addresses this for `[Default_Storage_Pool](https://docs.adacore.com/live/wave/arm22/html/arm22/arm22-13-11-3.html)`. `Standard` is an identifier specific to the pragma/aspect, and it is illegal if a declaration with that defining identifier is immediately visible at that point (see 4.1/4 and 5.1/4); anything else is resolved normally (5.2/5). The same rule could be transposed directly to `SPARK_Mode`.

thvnx:
> I'd prefer extending SPARK_Mode to accept static Boolean expressions alongside On and Off literals, rather than adding a second aspect and pragma.

Same. It also has the nice consequence of preserving `Auto`!

If @JoseRuizAdaCore's suggestion is not approved, the RFC needs to clarify how `SPARK_Mode_Enabled` and `SPARK_Mode` interact.

kanigsson:
Rewritten to hopefully match @JoseRuizAdaCore's suggestion; also enriched the "alternatives" section.

features/rfc-spark_mode-static-boolean.md
```diff
@@ -0,0 +1,137 @@
...
+
+Placement, inheritance and consistency rules are unchanged. The identifiers
+`On`, `Off` and `Auto` remain identifiers specific to the pragma/aspect, and
+keep their current meaning. Any other argument must be a static Boolean
+expression; `True` is equivalent to `On` and `False` is equivalent to `Off`.
+
+The names `On`, `Off` and `Auto` are resolved as follows, mirroring the rule
+that the ARM already uses for the `Default_Storage_Pool` aspect (ARM 13.11.3):
+an argument that is one of these identifiers is illegal if a declaration with
+that defining identifier is immediately visible at that point (13.11.3(4.1/4)
```
sttaft: I think "directly visible" would be better here, because you also want to disallow use of On/Off/Auto if entities with such a name are use-visible.  For "Standard" this wasn't an issue because Standard cannot be made use-visible (since package Standard would always hide such a declaration from being use visible).  But for On/Off/Auto, use-visibility is the most likely way that they will become directly visible, and the confusion would be very similar.

sttaft: For the rare case where this becomes a problem, True and False could be used instead of On/Off.  Auto is trickier, but presumably Auto is typically set at the compilation level, outside the scope of any "use" clauses.

sttaft: Other than shifting to "directly visible", the rest of the RFC looks good to me!

kanigsson: I have used "directly visible" now, plus added the explanation for this choice in the Rationale section, plus the workaround for On/Off clashes.

sttaft (approved):
Looks good to me!

kanigsson:
I have no clue about the RFC process. I know you guys have a meeting soon so I will leave this MR as-is for now until then. If on the contrary it is better to merge now for some reason, please go ahead.

tonunaks (approved):

PR #243, opened 2026-08-13
Generic actual inference

QuentinOchem:


features/rfc-inference-of-generic-actuals-from-call-actuals.md
```diff
@@ -127,6 +127,84 @@ Here:
 * The `"<"` is deduced as per pre-existing instantiation rules.
 
 
+Inference through generic packages
```
thvnx, line 130: This RFC is fundamentally a name-resolution change, so could you specify precisely how inference composes with overload resolution: in what order they run, which entities participate, and exactly which cases are considered ambiguous and rejected.

That precision matters for cost too: implementing this will be significant work in both GNAT and Libadalang, and their resolvers can only do it against well-defined name-resolution rules.

sttaft:
In the Prior Art section, you might want to talk about other languages that have done something similar.  I asked Google: "What languages support implicit instantiation as part of function call" and here is the response:

Programming languages that use generics or templates and perform type inference during a regular function call support implicit instantiation (automatically generating a specialized function for specific types based on the arguments provided). [[1](https://www.modernescpp.com/index.php/template-instantiation/), [2](https://www.learncpp.com/cpp-tutorial/function-template-instantiation/)]

Languages with Implicit Instantiation on Function Calls

- C++: The classic example. When you call a [C++ Function Template](https://en.cppreference.com/w/cpp/language/function_template) with concrete arguments (e.g., max(1, 2)), the compiler implicitly instantiates the matching specialization (max<int>). [[1](https://users.rust-lang.org/t/explicit-template-instantiation-in-rust/12891), [2](https://www.learncpp.com/cpp-tutorial/function-template-instantiation/)]
- Rust: Generic functions are implicitly instantiated (monomorphized) by the compiler when invoked with specific concrete types without requiring manual type parameters in the call syntax (e.g., some_function(5)infers <i32>).
- Swift: Uses generic functions where type parameters are implicitly deduced from function call arguments.
- D: Supports function templates that implicitly instantiate based on the types of arguments passed during a call.
- Scala (with generics/polymorphic methods): Type parameters on methods are inferred implicitly at the call site.

features/rfc-inference-of-generic-actuals-from-call-actuals.md
thvnx: I cannot push directly here so I opened https://github.com/AdaCore/ada-spark-rfcs/pull/251.

PR #244, opened 2026-08-20
[RFC] Tuples

ohenley:


features/rfc-tuples.md
```diff
@@ -0,0 +1,978 @@
...
+function result is, a function may return several values whose sizes it chooses
+itself — `{String; String}` — which no combination of `out` parameters can express.
+
+Motivation
+==========
+
+Returning more than one value from a function is one of the oldest ergonomic
+complaints about Ada, and every workaround costs something.
+
+`out` parameters force the operation to be a procedure, which excludes it from
```
sttaft, line 35: Ada allows functions to have OUT parameters, as of Ada 2012.

features/rfc-tuples.md
```diff
@@ -0,0 +1,978 @@
...
+     Key   : String (1 .. Key_Len);
+     Value : String (1 .. Value_Len);
+  end record;
+
+  function Split (Line : String; Sep : Character) return Split_Result;
+  ```
+
+- **`Unbounded_String`, or `Indefinite_Holders`,** move both results to the heap.
+  That rules the operation out of exactly the contexts Flare cares most about —
+  bare-metal and certified code, where the heap and the secondary stack are often
```
MartinClochardPro, line 108: I fail to see how the proposed solution may work without the secondary stack, which specifically addresses variable-length results.

features/rfc-tuples.md
```diff
@@ -0,0 +1,978 @@
...
+...
+Result := Divide (A, B);
+```
+
+A tuple is therefore best understood as a *deferred aggregate*: a composite value
+that has not yet been told which nominal type it belongs to. `{1, 2}` against a
+record target is already an aggregate of that record; this rule extends the same
+courtesy to a value whose type happens to be a tuple.
+
+Swapping
```
thvnx, line 382: I think swapping hides a few assumptions worth clarifying:
- it's only legal when A and B have the same type
- whether swapping limited A and B is legal

features/rfc-tuples.md
```diff
@@ -0,0 +1,978 @@
...
+   Ada generic code, and no such code can range over arity. Either the compiler
+   synthesizes `To_String` for tuples as it does `'Image`, or tuples are not
+   printable through the framework. This needs deciding with the `To_String`
+   owner.
+3. **Deep paths in component groups.** Flare's deep delta aggregates already allow
+   `A.B` on the left of an association, so `R.{A.B, C}` is a natural extension.
+   Not proposed here.
+4. **Array component groups.** `Arr.{1, 3} := {X, Y}` is conceivable but wants
+   reconciling with slice notation first. Not proposed here.
+5. **Arity 0 and 1.** `{X}` is unambiguous with braces and `{}` falls out of the
```
thvnx, line 873: Note: syntax rules already forbid Arity-0 and I would be in favor to forbid Arity-1 too.

features/rfc-tuples.md
```diff
@@ -0,0 +1,978 @@
...
+type Pair is {Integer; Integer};   --  illegal in this proposal
+```
+
+is not permitted. Naming a tuple type would immediately raise whether the name is
+a synonym or a distinct nominal type, which is a different feature — see *Future
+possibilities*. A subtype declaration is allowed, because a subtype never
+introduces a type and the synonym cannot disturb structural identity:
+
+```ada
+subtype Pair is {Integer; Integer};   --  legal: a synonym, same type
```
thvnx, line 456: I'm not convinced that `subtype` is a right choice here because for tuples it doesn't support contracts (no predicate, no constraint) which regular subtypes allows (here `{Integer, Integer}` is an anonymous type that can't be named I guess). That makes it hard to use it in SPARK if it cannot be used in contracts (unless decomposition is extended to support that case).

thvnx (commented):
Thanks @ohenley for the detailed RFC! I've only reviewed part of it so far. I'm wondering if the decomposition sections might deserve it's own RFC, splitting it out would make the whole thing easier to review (just a suggestion).

features/rfc-tuples.md
```diff
@@ -0,0 +1,978 @@
...
+
+Result : Division_Result;
+...
+Result := Divide (A, B);
+```
+
+A tuple is therefore best understood as a *deferred aggregate*: a composite value
+that has not yet been told which nominal type it belongs to. `{1, 2}` against a
+record target is already an aggregate of that record; this rule extends the same
+courtesy to a value whose type happens to be a tuple.
```
ronan-d, line 380: I think this part goes a little too far. Currently, when you see

```ada
Result : Division_Result := Divide (A, B);
```

you know that the author of `Divide` has had to read the definition of `Division_Result`.

With the kind of implicit conversions proposed here that would no longer be true. So I would be in favor of dropping this aspect of the proposal. I think that can be done without altering the "returning multiple values ergonomically" side. And so the "deferred aggregate" notion mentioned in the abstract would go away as well.

features/rfc-tuples.md
```diff
@@ -0,0 +1,978 @@
...
+
+Syntax
+------
+
+Note on notation: the Ada reference manual's grammar metasyntax already uses `{`
+and `}` for "zero or more". The wording will therefore need a convention for
+literal braces; below, literal braces are shown in bold as **{** and **}**.
+
+```
+tuple_definition ::= **{** tuple_component_list **}**
```
ronan-d, line 431: The surrounding code block breaks the "bold" markup here.

ronan-d, line 431: I would call this "tuple expression" rather than "tuple definition". "Definition" suggests that you can refer to it later using some name, which is not the case here.

features/rfc-tuples.md
```diff
@@ -0,0 +1,978 @@
...
+
+Reference-level explanation
+===========================
+
+Syntax
+------
+
+Note on notation: the Ada reference manual's grammar metasyntax already uses `{`
+and `}` for "zero or more". The wording will therefore need a convention for
+literal braces; below, literal braces are shown in bold as **{** and **}**.
```
sttaft, line 428: The Ada RM has already dealt with this problem for '[' and ']' and the solution is to explicitly quote metacharacters when they are part of the syntax of some construct.  So simply writing '{' and '}' should be sufficient.  See [Ada RM 4.3.3](https://ada-rapporteur-group.github.io/ARM/Ada_202Y/AA-4-3-3.html) for an example.

PR #246, opened 2026-08-20
[Flare] Initial draft for mandatory {} delimiters for record aggregates

### LINKED ISSUE #245: [feature] Flare: mandatory {} delimiters for record aggregates
This RFC makes a curly-brace syntax `{...}` mandatory for record aggregate
construction in pedantic Ada Flare.

joaopsazevedo:
Closes https://github.com/AdaCore/ada-spark-rfcs/issues/245

### DIFF
--- a/features/rfc-mandatory_curly_braces.md
+++ b/features/rfc-mandatory_curly_braces.md
@@ -1,0 +1,472 @@
+- Feature ID: mandatory_curly_braces
+- Start Date: 2026-08-13
+- Status: Design
+
+# Summary
+
+This RFC makes a curly-brace syntax `{...}` mandatory for record aggregate
+construction in pedantic Ada Flare:
+
+1. Record aggregates: `{X => 1, Y => 2}`, `{1, 2}`, `{others => <>}`, and `{}`
+   for the empty (null) record aggregate
+2. Extension aggregates: `{Base with {X => 1}}`, `{Base with {}}`
+3. Record delta aggregates: `{Base with delta {X => 1}}`
+
+In extension and delta aggregates, the components that follow `with` are
+themselves wrapped in braces, so what follows `with` always has the shape of a
+record aggregate; a null extension is written `{Base with {}}`.
+
+Because braces remove the collision between a record aggregate and a
+parenthesised expression, this RFC also lifts the Ada rule that a record
+aggregate with a single component association must be written in named notation:
+`{42}` is a legal positional record aggregate in pedantic Flare.
+
+# Motivation
+
+## Readability and visual clarity
+
+RFC `mandatory_square_brackets` removed most non-call uses of `()` from pedantic
+Flare, but record-shaped aggregates stayed parenthesised, so `()` still carries
+three meanings: call actual parameters `Make (X => 1)`, record aggregate
+`(X => 1, Y => 2)`, and parenthesised expression `(X)`. A record aggregate is
+textually identical to the actual parameter part of a call:
+
+```ada
+Draw (Point => (X => 1, Y => 2));   --  aggregate or inner call?
+Draw (Point => {X => 1, Y => 2});   --  unambiguously an aggregate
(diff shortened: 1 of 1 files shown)


features/rfc-mandatory_curly_braces.md
```diff
@@ -0,0 +1,472 @@
...
+Braced forms still require normal Ada resolution to distinguish a record
+aggregate from an extension aggregate and to resolve component names, but the
+construct family is determined by the delimiter alone.
+
+# Rationale and alternatives
+
+Curly braces are the natural choice for record construction:
+
+- They carry exactly the intended meaning in the languages most programmers come
+  from: struct and object construction in C, C++, Rust, Go, C#, Java,
```
thvnx, line 363: I think Java only uses curly braces for arrays.

features/rfc-mandatory_curly_braces.md
```diff
@@ -0,0 +1,472 @@
...
+## Lexical elements
+
+Curly braces are not delimiters in Ada. RM 2.2 must be extended so that `{` and
+`}` are single delimiters, alongside `(`, `)`, `[`, and `]`; no compound
+delimiter involving braces is introduced. This is a pure addition: `{` and `}`
+occur in legal Ada source only inside string literals, character literals, and
+comments, where they remain uninterpreted.
+
+RFC `string_interpolation` gives braces meaning inside interpolated string
+literals (`f"..."`). There is no grammar-level conflict but since an
```
thvnx, line 228: That's maybe not entirely true. In theory, I agree that there is no conflict, a parser should be able to handle this. In practice both GNAT and Libadalang parsers would need to be reworked since they can't support this construct yet (GNAT uses a simple boolean to track interpolated string expression, Libadalang uses a regexp, `{` and `}` are not even proper tokens in the grammar yet). So both frontend need changes in that regard.

thvnx (commented):
Thanks @joaopsazevedo for this really well written RFC! Looks good to me. My only concern is the `{}` vs `null record` notation case. On one hand, it mirrors empty array notation nicely, but on the other, it gets hard to read once nested. Keeping `null record` might actually be a better choice for the reader here.

features/rfc-mandatory_curly_braces.md
```diff
@@ -0,0 +1,472 @@
...
+`with` a uniform right-hand side: whatever follows it (after `delta`, if any) is
+a braced, record-shaped component list.
+
+## Record delta aggregates
+
+RM 4.3.4 `record_delta_aggregate` becomes:
+
+```
+record_delta_aggregate ::=
+  '{' base_expression with delta '{' record_component_association_list '}' '}'
```
MartinClochardPro, line 295: What is the motivation for introducing additional inner curly braces here ? I fail to see anything beyond consistency with extension aggregates. This feels a bit weak as extension aggregates are a highly specific corner case with their own specific syntax, and the change was forced by syntactic ambiguities
for extension aggregates rather than being a desirable esthetic decision. On the other hand,
I have several arguments against the change:

- Having additional braces creates a discrepancy with array delta aggregates (visible in the RFC text itself), which do not have inner brackets. The deep delta aggregate extension make clear that record delta and array delta are essentially the same feature; I would expect consistency to be more important here than between record delta and record extension.
- As stated in following lines, 'no empty form is provided', and 'The inner list is [...] named-only'. Either restriction prevent the ambiguity concerns that forced the introduction of these extra braces in extension aggregates in the first place, so nothing force a change.
- The additional braces do not match anything like a record, while these in the extension aggregate are in line with the syntax of the extension record type declaration (Aggregate `{ P'(X) with { F => ...; G => ...; ... }}` for `type T is new P with record F : ...; G : ...; ... end record`, inner braces match 'record  ... end record', and the language rules for the part in the inner curly braces are exactly the same as that for an aggregate for a type that would be declared by the record block).

PR #247, opened 2026-08-20
replaced 'Fixed by with Pinned (list), following last live discussions

QuentinOchem:


features/rfc-oop-first-controlling.rst
```diff
@@ -141,25 +151,29 @@ between the partial and the full view:
    type T is tagged null record with First_Controlling_Parameter; -- ILLEGAL
 
 
-'Fixed and renamings
+Pinned and renamings
 --------------------
 
-'Fixed participate to subprogram conformance, and need to match when creating
-renamings. Notably, in the following example:
+``Pinned`` participates to subprogram conformance, and needs to match when
```
thvnx, line 157: Why not simply forbid `Pinned` on renaming declarations? Since spec/body, access to subprograms, or generic formals are not required to be conformant either, we can just treat it as a "declaration" aspect like `Pre`/`Post`. I would propose the following rewording:

> Pinned is specified by an aspect_specification on the primitive's declaration and is not part of its profile; like Pre/Post, it does not participate in subprogram conformance.

features/rfc-oop-first-controlling.rst
```diff
@@ -103,19 +103,29 @@ For example:
...
+    function F3 (V : Child) return Child with Pinned => (F3'Result);
+    -- Primitive, controlling on the first parameter only; the result is pinned
 
 As long as the first parameter is controlling, a primitive may also have
 additional controlling parameters and a controlling result. These all dispatch
 and are rewritten to the derived type when the primitive is overridden. When a
-parameter or result should refer to the type without dispatching, it is marked
-with the ``'Fixed`` attribute, as shown above with ``F3``.
+parameter or result should refer to the type without dispatching, it is listed
+in the ``Pinned`` aspect of the subprogram, as shown above with ``F3``.
```
thvnx, line 113: The Pinned rules don't mention `First_Controlling_Parameter`, could you state explicitly whether `Pinned` requires `First_Controlling_Parameter` or works on any tagged type's primitive?

QuentinOchem: There's no reason that I can think of to limit with Pinned to First_Controlling_Parameter - so it should be available to all tagged types primitives indeed.

features/rfc-oop-first-controlling.rst
MartinClochardPro: It is a bit more complicated than that; the primitiveness of two views of the same subprogram may actually  differ. See [ARM 8.5.4 (12)](http://ada-auth.org/standards/22rm_w_amd1/html/RM-8-5-4.html).

> The primitiveness of a renaming-as-declaration is determined by its profile, and by where it occurs, as for any declaration of (a view of) a subprogram; primitiveness is not determined by the renamed view. In order to perform a dispatching call, the subprogram name has to denote a primitive subprogram, not a nonprimitive renaming of a primitive subprogram

In particular, a primitive subprogram renaming(-as-declaration) another create two independently overridable primitives, as you can see by running the example below. Meaning that in fact, 'Pinned for a renaming is pertinent, and should be totally unrelated to 'Pinned on the renamed subprogram.

(I am not defending the current situation, this is highly counter-intuitive, completely at odds with the expectation that a renaming should just be a short-hand for the same thing. Which is pretty much why GNATprove have been rejecting primitive renaming-as-declaration since the earliest version.)

```ada
with Ada.Text_IO; use Ada.Text_IO;
procedure Main with SPARK_Mode is
   package A is
      type T is tagged null record;
      procedure P (X : T);
      procedure Q (X : T) renames P;
   end A;
   package body A is
      procedure P (X : T) is begin Put_Line ("Hello World !"); end P;
   end A;
   use A;
   package B is
      type U is new T with null record;
      overriding procedure P (X : U);
      overriding procedure Q (X : U);
   end B;
   package body B is
      procedure P (X : U) is begin Put_Line ("Hello.... "); end P;
      procedure Q (X : U) is begin Put_Line (" ....World !"); end Q;
   end B;
   use B;
   X : T'Class := U'(null record);
begin
   P (X); Q (X); -- Prints "Hello.... \n ....World !"
end Main;```

PR #251, opened 2026-09-01
Complete the generic inference RFC reference-level section

thvnx:
I've completed the reference-level section of the RFC on inferring generic actuals from call actuals. I derived everything from the related RFCs and existing Ada rules, with AI assistance, so it probably deserves a careful review.

I also fixed the TODO about the mutable state for the Allow_Structural_Instantiation aspect.


sttaft:
I think it makes more sense to first try to infer an actual, and only use the default if there is not enough information to infer the actual.  In general, I would expect that inference only happens when there is only one possible actual that works.  That would mean the default is either the same as what is inferred, or in fact wouldn't work.  So using it when there is something to be inferred seems to never be the right choice.

thvnx:
> I think it makes more sense to first try to infer an actual, and only use the default if there is not enough information to infer the actual.

Agreed, and updated the RFC!

PR #254, opened 2026-09-09
[RFC] register access model

JoseRuizAdaCore:
Add who and when components in a hardware register may change.


sttaft:
Looks very interesting!  One question is whether it might make sense to combine multiple aspects into a single composite aspect, in the way that was done for the Aggregate aspect.  This might give more flexibility in the future, where the various components of the composite aspect could be revised or extended without the danger of bumping into aspect names that other vendors might have chosen for other purposes.

JoseRuizAdaCore:
Thanks Tuck!

We could merge the *who* (`Program_Access`, `Environment_Writes`, and `Reserved`) under something like `Accessed_By`, together with a new aspect for reserved components (`Write_Value`):
```
   --  "The prescaler is set by software and sampled by the timer.
   --   It is never modified by hardware."
   Prescaler  : Bits_4  with Accessed_By => (Program     => Read_Write,
                                             Environment => Read);

   --  "Writing a one starts the counter.  The bit reads back the run
   --   state, which hardware clears at top-of-count in one-shot mode."
   Running    : Boolean with Accessed_By => (Program     => Read_Write,
                                             Environment => Read_Write);

   --  "Reserved.  Must be written as zero."
   Reserved_1 : Bits_3  with Accessed_By => (Program => None),
                             Write_Value => 0;

   --  "Reserved.  Software must preserve the value read."
   Reserved_2 : Bits_3  with Accessed_By => (Program     => None,
                                             Environment => Write),
                             Write_Value => Preserve;
```

We could merge the *effect* aspects (`Write_Effect` and `Read_Effect`) under `Effect`:
```
   --  "Reading returns the oldest byte and advances the read pointer.
   --   Writing appends a byte.  The field itself holds no state."
   Data    : Byte   with Effect => (Read  => External,
                                    Write => (Trigger_On_Write => External));
```

And we could merge the *when* (`Write_When` and `Read_When`) aspects under `Access_When`:
```
type UART_Peripheral is record
   CR : Control_Register;   --  Enable, Prescaler, Baud_Rate
   SR : Status_Register;    --  RX_Ready, TX_Empty, Busy
   DR : Data_Register;
end record
  with Volatile,
       Access_When =>

         (Write => (--  "The baud divisor may only be changed while the
                    --   UART is disabled."
                    CR.Baud_Rate when not CR.Enable,

                    --  "No control field may be written while a
                    --   transmission is in progress."
                    CR           when not SR.Busy),

          Read  => (--  "DR may only be read when RX_Ready is set;
                    --   otherwise the value is undefined and the
                    --   receiver stalls."
                    DR           when SR.RX_Ready));
```

I think it would be worth overall.

sttaft:
Looks great!

clairedross:
It does seem complex, and, at this stage, it is unclear to me whether GNATprove can ensure the safety of all this. Since this RFC seems to be heavily proof-oriented (right?), I think it is something that we need to make sure of. I think we might need a working group - at least a SPARK team member and Jose - to take a look.

JoseRuizAdaCore:
I think it may be split into two/three smaller ones. Let me think about it.

